PR #31, opened 2014-12-17
Allow values to insert to be passed as array of setters

kasrak:
Hello!

It can be helpful to build up the setters for a query dynamically. If Swift supported splatting, then the existing `insert` could be used (e.g. `insert(*[name <- "Alice", email <- "alice@mac.com"])`), but it doesn't (https://devforums.apple.com/message/970958).

This diff lets the columns be built up dynamically. For example:

    var values = [name <- "Alice"]

    if shouldSetEmail {
        values.append(email <- "alice@mac.com")
    }

    if let insertedID = users.insert(values) {
        ...
    }

If you think this is a good idea, `update` should also be changed to allow an array of setters.

### DIFF
--- a/SQLite Common/Query.swift
+++ b/SQLite Common/Query.swift
@@ -449,7 +449,19 @@ public struct Query {
         return insert([value] + more)
     }
 
-    private func insert(values: [Setter]) -> (ID: Int?, statement: Statement) {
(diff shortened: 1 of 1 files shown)


stephencelis:
Thanks! Sorry for the delay!

PR #43, opened 2015-01-18
Update Index.md

ericcgu:


stephencelis:
Good idea! More visual aids are always a good thing. I think we'll have to make sure the image is part of the repo itself, though, rather than a comment's image on GitHub's CDN. I'll try to add an image soon.

ericcgu:
Thanks Stephen. I know i had to spent about two minutes figure it out
before and thats two minutes saved for someone else.

stephencelis:
Closing in favor of https://github.com/stephencelis/SQLite.swift/commit/3d7ad5dfb8c6590355b6a92556ae47ffca8addaf

@ericcgu Just a note, you shouldn't need to muck around in the build phases tab for resources if you make sure to check **Copy items if needed** and your app under **Add to targets:** when dragging the database into your Xcode project. You can also select the file from the left-hand file inspector, and use the **Target Membership** section of the right-hand inspector to manage which targets bundle the resource.

PR #45, opened 2015-01-20
Add transaction method which takes an array of Statements as its unique parameter

juliensagot:


stephencelis:
Do you have more information on the problem you're solving for here? Is this a matter of avoiding calling `.Deferred` explicitly?

stephencelis:
Ah, never mind. I didn't realize that the array function was private. Should we publicize that one, too?

juliensagot:
Yeah, that would be great ! 👍

stephencelis:
Thanks!

PR #62, opened 2015-02-10
Swift 1.2

stephencelis:
This PR is open to track changes against the Swift 1.2 beta. It won't be merged till Swift 1.2 goes GM, but will be continuously rebased onto `master`.

Most things should work as they have been, with a few notes:

  - The `db.transaction` helpers that took variadic auto-closures have been removed (Swift no longer supports variadic auto-closures). Update path: use `&&` and `||` for control flow:
    ``` swift
    db.transaction() &&
        stmt1 &&
        stmt2 &&
        db.commit() || db.rollback()
    ```
    Or use the block-based helper:
    ``` swift
    db.transaction { _ in
        stmt1.run()
        if stmt1.failed { return .Rollback }
        stmt2.run()
        if stmt2.failed { return .Rollback }
        return .Commit
    }
    ```
    > _Note:_ You'll need to explicitly call/return COMMIT and ROLLBACK now.

  - ~~There appears to be a bug in Swift causing 2 crashes in the test suite. Filed: rdar://19782170~~ _Fixed in beta 2._

#### TODO:

  - [x] Update `db.transaction` helpers.
  - [x] Update `db.savepoint` helpers.
  - [ ] Update README.
  - [ ] Update documentation.
  - [ ] Update playground.

The tests have also been heavily refactored (they were abusing the power of `@autoclosure`, which has been curtailed with `@noescape`), but should be generally more readable, if slightly less flexible. _Beta 2 update: `@autoclosure(escaping)` is now an option._


iancanderson:
Just wanted to say thanks for this awesome library and I'm cheering you on to keep up with the Swift releases :+1: :smile:

stephencelis:
Changes have been slowing down so hopefully the upgrade will go smoothly :blush:

PR #64, opened 2015-02-13
sqlcipher support

biohazardffm:
- compile using sqlcipher repository
- add key and rekey functions; rekey(0,0) does not work yet as I could not figure out the right type in swift

### DIFF
--- a/SQLite Common/Database.swift
+++ b/SQLite Common/Database.swift
@@ -405,9 +405,7 @@ public final class Database {
     }
 
     internal func try(block: @autoclosure () -> Int32) {
-        perform {
-			if block() != SQLITE_OK { assertionFailure("\(self.lastError)") }
-		}
+        perform { if block() != SQLITE_OK { assertionFailure("\(self.lastError)") } }
     }
 
     // MARK: - Threading
--- a/SQLite Common/Database.swift
+++ b/SQLite Common/Database.swift
@@ -51,6 +51,25 @@ public final class Database {
 
     deinit { try(sqlite3_close(handle)) } // sqlite3_close_v2 in Yosemite/iOS 8?
 
+	// MARK: - sqlcipher support
+	
+	/// Specify the key for an encrypted database. Should be called right after Database().
+	/// If the current database is not encrypted, this routine will encrypt it.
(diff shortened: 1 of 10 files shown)


stephencelis:
Nice! Thanks for the PR!

Is there any way to compose this so it's opt-in given a build flag, and without SQLCipher it uses the OS-bundled libraries?

biohazardffm:
Welcome! Not sure but I think it should be possible. Tried to do an extension on Database before in a different framework but that did not work....
Still learning swift... Will try to add a build flag and get back to you.

SQLite Common/Database.swift
```diff
@@ -51,6 +51,25 @@ public final class Database {
...
+	public func key(key:String) {
+		try(sqlite3_key(handle, key, Int32(key.lengthOfBytesUsingEncoding(NSUTF8StringEncoding))))
+	}
+
+	/// Change the key on an open database. If the current database is not
+	/// encrypted, this routine will encrypt it.
+	/// If key == "" the database is decrypted.
+	public func rekey(key:String) {
+		if(key.lengthOfBytesUsingEncoding(NSUTF8StringEncoding) > 0) {
+			try(sqlite3_rekey(handle, key, Int32(key.lengthOfBytesUsingEncoding(NSUTF8StringEncoding))))
```
stephencelis: You should be able to keep the byte length Swift-native by calling:

``` swift
countElements(key.utf8)
```

biohazardffm: works, but still have to wrap it in an Int32()

biohazardffm:
I have added 2 new sqlcipher enabled targets. Will add some tests and then it should be good to go.

stephencelis:
Thanks for your work on this! Merged it and rebased my universal framework branch against it. Should simplify things by only requiring two targets.

manojshrestha:
HI Stephencelis,
I was trying to add SQLCipher in my project but there is error saying "No such module SQLiteCipher". I have added it in both Embed frameworks and Target Dependencies. (Manual Installation)

It would be helpful if you add runnable example project with the code.
Thank you,

PR #76, opened 2015-03-04
Support non-integer primary key in shorthand function

xenadu:
Rather than rely on the documentation, I figure why not support this directly? 

It has the nice benefit of making the generated SQL syntax match what you'd write by hand, though it doesn't really change the functionality.

Included test of `String` primary key and custom type primary key.

### DIFF
--- a/SQLite Tests/TestHelper.swift
+++ b/SQLite Tests/TestHelper.swift
@@ -7,6 +7,7 @@ let age = Expression<Int?>("age")
 let salary = Expression<Double>("salary")
 let admin = Expression<Bool>("admin")
 let manager_id = Expression<Int64?>("manager_id")
+let uniqueIdentifier = Expression<UniqueIdentifier>("uniqueIdentifier")
 
 class SQLiteTestCase: XCTestCase {
 
@@ -70,4 +71,17 @@ class SQLiteTestCase: XCTestCase {
         if let count = trace[SQL] { trace[SQL] = count - 1 }
     }
 
+}
+
+
+public class UniqueIdentifier : NSUUID, Value {
+    public static var declaredDatatype = "TEXT"
+    
+    public static func fromDatatypeValue(datatypeValue: String) -> UniqueIdentifier {
+        return UniqueIdentifier(UUIDString: datatypeValue)!
+    }
+    
+    public var datatypeValue: String {
+        return self.UUIDString
+    }
 }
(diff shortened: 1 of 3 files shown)


stephencelis:
> It has the nice benefit of making the generated SQL syntax match what you'd write by hand, though it doesn't really change the functionality.

Using the column constraint actually does still match (and is required for composite keys, for example).

This was actually originally a design decision. The assumption being that integer primary keys were the most common and to avoid the mistake of passing in just anything. This makes less sense now in hindsight, though, especially given that `.Autoincrement` is supported.

Let me check this out in the evening to make sure it doesn't create any ambiguity depending on how the functions are called. Leaving a few minor style comments in the meantime.

SQLite Tests/TestHelper.swift
```diff
@@ -71,3 +72,16 @@ class SQLiteTestCase: XCTestCase {
     }
 
 }
+
+
+public class UniqueIdentifier : NSUUID, Value {
```
stephencelis: I know this is just test, but any reason you're not extending NSUUID directly?

SQLite/Schema.swift
```diff
@@ -193,20 +193,20 @@ public final class SchemaBuilder {
...
-    public func column<V: Value where V.Datatype == Int64>(
+    
+    public func column<V: Value>(
         name: Expression<V>,
         primaryKey: Bool,
         unique: Bool = false,
         check: Expression<Bool>? = nil
     ) {
-        column(name, primaryKey ? .Default : nil, false, unique, check, nil, nil)
+            column(name, primaryKey ? .Default : nil, false, unique, check, nil, nil)
```
stephencelis: Looks like an extra tab slipped in here.

stephencelis:
Hey, I'd love to merge this in but it's off the `swift-1-2` branch instead of `master`. Do you think you could re-open it against master, instead? Or would you rather me commit the changes myself and close this when I'm done?

SQLite Tests/TestHelper.swift
xenadu: Only because I already had the `UniqueIdentifier` class definition handy and it tests that there are no problems using a completely custom column datatype as the primary key.

I did a test that uses the email field which is a `String` to cover that case.

SQLite/Schema.swift
xenadu: Xcode auto-format on paste, heh. I'll fix it.

xenadu:
@stephencelis My assumption was that `swift-1-2` will merge into or replace master when the next version releases. How are you handling merging changes into `swift-1-2`?

Next week I'm going to try to submit a PR to add a few things to `Database` like WAL mode checkpointing, but I'm only using the `swift-1-2` branch since I'm doing my new app on the beta

stephencelis:
Closing in favor of #78. Will rebase `swift-1-2` soon.

PR #78, opened 2015-03-06
Support non-integer primary key in shorthand function

xenadu:
Opened a version of #76 against master (and swift 1.1). Feel free to merge either


stephencelis:
Thanks!

stephencelis:
FYI, I've also enabled `defaultValue` for primary keys, that way you can automate UUID generation:

``` swift
let uuid = Expression<String>("uuid")
let uuidgen: () -> Expression<String> = db.create(function: "uuidgen") {
    return NSUUID().UUIDString
}

db.create(table: users) { t in
    t.column(uuid, primaryKey: true, defaultValue: uuidgen())
}
```

xenadu:
Oh nice 👍

Issue #77, opened 2015-03-04
Support for Carthage


JeffBNimble:
I'm trying to reference your project using Carthage and it won't resolve since you have no releases. Would you mind creating a release, even if it's something like 0.0.1?

Carthage apparently will not resolve anything that doesn't have a release. See [https://github.com/Carthage/Carthage](here) for more info.

stephencelis:
Carthage shouldn't require a version. You simply need to specify a branch.

```
github "stephencelis/SQLite.swift", "swift-1-2"
```

**However**, SQLite.swift can't be built using Carthage right now, and I'm not sure I'll ever be able to support it. There are two blockers:

- https://github.com/Carthage/Carthage/issues/340
- https://github.com/Carthage/Carthage/issues/341

Beyond this, I think Carthage's pre-compiled frameworks aren't compiled with the same build settings as your project, so things that require `DEBUG` (like assertions) wouldn't work as expected in development.

I highly recommend using the installation instructions given in the README, for now, and checking SQLite.swift as a submodule that integrates into Xcode as a working copy.

stephencelis:
The `wontfix` label is tentative. I'll explore moving back to a non-universal framework sometime in the future, though I'd want to thoroughly test that assertions are properly surfaced. If not, I'll have to also revisit how assertions are raised.

stephencelis:
@ctava Sorry, I missed your comment earlier. If you have a proposed PR fix, please submit it, but at the moment the fixes are needed in Carthage itself. I'd love to have SQLite.swift work out-of-the-box for both Carthage and CocoaPods, but at the moment its setup seems to prohibit it.

stephencelis:
Going to reopen this with a `help wanted` tag. I'd prefer a pull request without compromises in the complexity of how SQLite.swift is structured as a project and repository, but will make compromises where it makes sense.

Rupert-RR:
Looks like https://github.com/Carthage/Carthage/pull/622 fixes part of this!

schickling:
Yes with the next Carthage version universal builds are supported finally! :+1:

stephencelis:
Closing this as support will come with Carthage!

lukescott:
Should probably keep this open until support for universal builds are released. The current release, 0.8, does not include this.

For others that need this now, you have to install the latest version on master by running:
```bash
brew remove carthage
cd ~/Desktop
git clone https://github.com/Carthage/Carthage
cd ./Carthage
make install
```

[...]

PR #124, opened 2015-05-04
Added iOS Framework target for Carthage support

nvh:
This creates a separate iOS target that can be picked up by Carthage. This enables iOS support for Carthage out of the box. I've read #77 and understand that this is working around problems that should be fixed in Carthage, but I think this this little bit of overhead is acceptable.
Please let me know if you agree and this if this is a solution you would consider merging, then I could do the same for the Cipher target (this fork is basically scratching my own itch, but I think merging this helps a more widespread adoption of the project)

### DIFF
--- a/SQLite-iOS.plist
+++ b/SQLite-iOS.plist
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>en</string>
+	<key>CFBundleExecutable</key>
+	<string>$(EXECUTABLE_NAME)</string>
+	<key>CFBundleIdentifier</key>
+	<string>com.stephencelis.$(PRODUCT_NAME:rfc1034identifier)</string>
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleName</key>
+	<string>$(PRODUCT_NAME)</string>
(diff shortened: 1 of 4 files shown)


stephencelis:
Thanks for trying to tackle this. Looks like it doesn't address the Mac build and SQLCipher builds, though? (Which I'm not sure can't be done without tags or without requiring a different framework/module name).

I worry that Carthage builds frameworks in **Release** mode even when your app target is in **Debug** mode (please correct me if I'm wrong). SQLite.swift relies on Swift assertions for a lot of developer-related debugging, and if these get stripped out it would make SQLite.swift tough to troubleshoot.

nvh:
Nope, it doesn't yet, but I can add them in a similar way, but the SQLCipher version of the framework will get a different name.

I will look into the Release/Debug mode thing. Carthage separate the build step of dependencies from the building of the project, so it probably isn't possible do this automatically, but maybe you can tell carthage to build the frameworks in Debug mode as well.

baryon:
Nope, Waitting

stephencelis:
@baryon You can preliminarily try the branch referenced in this PR. I do want to have Carthage support but it requires a bit of thought and project restructuring that I haven't had time for yet.

kylejm:
@stephencelis any updates on this? I'd love to be able to add SQLite.swift via Carthage. What doesn't work about @nvh's solution to Debug builds?

stephencelis:
@kylejm A restructuring and decoupling on the Swift 2 branch should make Carthage support possible (see #154). I'm not sure how easy it will be to back-port to the current Swift 1.2 branch, though.

schickling:
I was able to install SQLite (swift-2 branch) via Carthage. :+1:

stephencelis:
Carthage supports universal frameworks now, and the `swift-2` branch has been updated to have multiple targets, so this should be good to close. Thanks!

PR #150, opened 2015-07-13
Added getter for query to return private columnNames variable

Hexidecibel:
When iterating through a result set obtained via a select * query, the exact number of columns is unknown.  The columnNames variable is needed to properly iterate through this kind of result set and this commit provides access to it via the Query object.

    var aliasByColumnName = [String: String]
    var aliases = database("SELECT * FROM ALIASES")
    for alias in aliases {
      var columnNames:[String: Int] = alias.getColumnNames()
      for (key, value) in columnNames {
        aliasByColumnName[key] = alias.values[value]
      }
    }

### DIFF
--- a/SQLite/Query.swift
+++ b/SQLite/Query.swift
@@ -658,6 +658,11 @@ public struct Row {
     private init(_ columnNames: [String: Int], _ values: [Binding?]) {
         (self.columnNames, self.values) = (columnNames, values)
     }
+    
+    /// Returns columnNames variable - for mapping select * queries
+    public func getColumnNames() -> [String: Int] {
+        return self.columnNames
+    }
 
     /// Returns a row’s value for the given column.
     ///


stephencelis:
Hm, exposing this in the type-safe layer doesn't quite work because the `Row` subscript takes an `Expression`. Did you mean to do this in `Statement`?

This does expose a gap in iteration, though, and perhaps the generator needs to iterate in a `(key, value)` fashion:

``` swift
for (column, value) in aliases.select(*) {
    …
}
```

Hexidecibel:
Woops, yes.  It worked for my purposes when in query but it does belong better in statement.  

I agree -- adding key/value iteration to the generator would satisfy my needs - and might make more sense from a usability standpoint.

stephencelis:
I'll keep this open and try to address it more generally soon. Thanks for the catch!

stephencelis:
Ah, I must have publicized this awhile ago and forgot to close the pull. Thanks!

PR #153, opened 2015-07-15
Add select functions with array param for query builder

huynhphan89:


stephencelis:
Thanks! The interface in general will provide array calling for vararg functions in the upcoming Swift 2 version.

PR #172, opened 2015-08-08
SQLiteCipher NSData key support

agisboye:
Adds key and rekey methods that accept a key of type NSData.
This allows the database to be easily keyed with a key generated by
Security.framework:
```swift
let keyData = NSMutableData(length: 64)!
let result = SecRandomCopyBytes(kSecRandomDefault, 64,
UnsafeMutablePointer<UInt8>(keyData.mutableBytes))

db.key(keyData)
```

### DIFF
--- a/SQLite/Value.swift
+++ b/SQLite/Value.swift
@@ -62,6 +62,10 @@ public struct Blob {
     public init(bytes: UnsafePointer<Void>, length: Int) {
         data = NSData(bytes: bytes, length: length)
     }
+    
+    public init (data: NSData) {
+        self.data = data
+    }
 
 }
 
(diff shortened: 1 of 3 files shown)


stephencelis:
Thanks for the pull request! A few small things and I'd love to merge this.

- I've been trying to segregate Foundation code wherever possible to keep SQLite.swift as portable as possible (especially since Swift will be on Linux in the future, but Foundation may not be). Any chance you can use `SQLite.Blob`, instead?

- Would you please maintain the whitespace of the original file? It looks like your lines use tabs, but the rest of the project uses 4 spaces.

- Would you write a quick couple tests for the cipher target? They should basically match the other two tests.

stephencelis:
@agisboye Any update here? If you don't want to make the changes I'll try to implement them when I have the chance.

agisboye:
@stephencelis Sorry for the delay. I've switched to SQLite.Blob and fixed  the whitespace. I haven't gotten around to the tests yet I'm afraid.

stephencelis:
Looking good! Let me know if you need any direction getting some tests written!

SQLite/Value.swift
```diff
@@ -62,6 +62,10 @@ public struct Blob {
     public init(bytes: UnsafePointer<Void>, length: Int) {
         data = NSData(bytes: bytes, length: length)
     }
+    
+    public init (data: NSData) {
```
stephencelis: Any chance we can get rid of this convenience initializer? The `swift-2` branch actually uses a fully Swift implementation of `Blob` (no reliance on `Foundation`/`NSData`), which I may back-port, at which point you can still create a blob from `NSData` like this:

``` swift
// data: NSData
let blob = Blob(bytes: data.bytes, length: data.length)
```

stephencelis:
Any update here? Would love to merge this if you have time to add some tests!

agisboye:
I've added tests, but I had to submit a new pull request (#317). Go ahead and remove this PR (#172).

stephencelis:
Closing in favor of #317.

PR #181, opened 2015-08-23
String and String.characters no longer Sliceable

nigelkerr:
Xcode-Beta Version 7.0 beta 4 (7A165t) on 2015-08-22 did not think String.characters had a split method.  this .componentsSeparatedByString(String) call appears to do what we needed to do, split the .expression.template on '.'

As an aside, this change allowed this same Xcode/Swift2 version and Carthage 0.7.5 to build all four schemes of SQLite.swift/swift-2 with this Cartfile entry:

github "stephencelis/SQLite.swift" "swift-2"

Though it must be said I've only used the SQLite iOS scheme results.

HTH!  Thank you for this framework!

### DIFF
--- a/Source/Typed/Query.swift
+++ b/Source/Typed/Query.swift
@@ -849,7 +849,7 @@ extension Connection {
         let columnNames: [String: Int] = {
             var (columnNames, idx) = ([String: Int](), 0)
             column: for each in query.clauses.select.columns ?? [Expression<Void>(literal: "*")] {
-                var names = each.expression.template.characters.split { $0 == "." }.map(String.init)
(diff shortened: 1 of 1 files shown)


stephencelis:
The `swift-2` branch is intended to be used with the latest version of Xcode 7 (beta 5). Can you please upgrade and try again?

nigelkerr:
Sorry for the noise: HEAD of swift-2 builds just fine with beta5, which I had not noticed.

Still, thank you for this framework!

stephencelis:
Glad to hear it solved your problem!

PR #182, opened 2015-08-24
fix podspec var name

toddkrabach:
fix the podspec for the swift-2.0 branch


stephencelis:
Ah, thanks! I'll try to keep this rebased as I continue to force-push the branch.

PR #185, opened 2015-08-25
XCode 7.0 Beta 6 Swift 2.0 conversion

fluffyemily:
fixed issues as best I could with current knowledge re: Xcode 7 beta 6 language changes

Converted to Swift 2.0 for XCode 7.0 Beta 6. Compiles but does not pass tests.

### DIFF
--- a/Source/Extensions/FTS4.swift
+++ b/Source/Extensions/FTS4.swift
@@ -101,12 +101,12 @@ public struct Tokenizer {
         }
 
         if !tokenchars.isEmpty {
-            let joined = "".join(tokenchars.map { String($0) })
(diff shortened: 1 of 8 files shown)


stephencelis:
Yeesh, the crashes are deeper than I thought. Looks like memory storage in Swift doesn't map as directly to Objective-C as before. Slowly converting things over to use `unsafeBitCast` and the Swift 2 C function pointer API.

stephencelis:
Merged your commit and applied fixes for the function pointer issues. Thanks again!

fluffyemily:
thank you so much :+1:

PR #189, opened 2015-08-27
Change Blob internal value from NSData to [UInt8]

### LINKED ISSUE #188: Back Blob with [UInt8] instead of NSData?
Can Blob be backed by [UInt8] instead of NSData? It would be one less Foundation dependency. Something like:
```swift
public struct Blob {
	let data: [UInt8]

	public init(data: [UInt8]) {
		self.data = data
	}

	public init(bytes: UnsafePointer<Void>, length: Int) {
		data = [UInt8](UnsafeBufferPointer<UInt8>(
			start: UnsafePointer<UInt8>(bytes),
			count: length
		))
	}
}
// deprecate (backwards compatibility)
extension Blob {
	public var bytes: UnsafePointer<Void> {
		return UnsafePointer<Void>(data)
	}

	public var length: Int {
		return data.count
	}
}
```

lukescott:
Fixes #188

### DIFF
--- a/SQLite/Value.swift
+++ b/SQLite/Value.swift
@@ -22,8 +22,6 @@
 // THE SOFTWARE.
 //
 
-import Foundation
-
 /// Binding is a protocol that SQLite.swift uses internally to directly map
 /// SQLite types to Swift types.
 ///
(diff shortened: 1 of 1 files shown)


stephencelis:
Thanks!

PR #190, opened 2015-08-27
Replace Foundation String format with Swift version

lukescott:
Missed a Foundation dependency that is automatically bridged: `String(format:)`. Some reason it still works without the import.

Added toHex() method to Blob and replaced `String(format:)` in Blob.description with it. Added a test for it.


stephencelis:
Awesome! Thanks!

PR #195, opened 2015-09-07
Add an initializer that takes an sqlite3 handle

bwoods:
At the moment SQLite.swift can’t be used on `sqlite3 *`s that it didn’t create. In my case, I have a database pre-configured with a C++ [VFS](http://www.sqlite.org/vfs.html) and an `sqlite3_update_hook `.

This pull request includes the minimal change that allow SQLite.swift’s type-safety (and other goodness) to be used with an `sqlite3 *` that has already been opened.

If you can think of a better approach, I’m happy to change the code…

### DIFF
--- a/SQLite/Database.swift
+++ b/SQLite/Database.swift
@@ -48,6 +48,7 @@ public final class Database {
     }
 
     internal var handle: COpaquePointer = nil
+	 private var owned: Bool = true
 
     /// Whether or not the database was opened in a read-only state.
     public var readonly: Bool { return sqlite3_db_readonly(handle, nil) == 1 }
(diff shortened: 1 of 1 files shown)


stephencelis:
Thanks for the PR! I definitely think that the current API would be improved with public access to the handle, including an initializer. Let me think about the interface a wee bit more. I'll try to merge this with a few changes for flexibility when I can.

bwoods:
Since Xcode 7 is now released, I have (locally) reproduced this commit on the `swift-2` branch. Should I update the pull request or just leave it as is? Which would be easier for you?

bwoods:
I now include SQLite.swift’s sources directly in my own framework, giving me (internal) access to the `handle`. As such, this PR is no longer needed; by me at least.

stephencelis:
Going to close this one out for now. I like the idea but the lower-level API needs some thought before integrating better with the C libs. Thanks again, though!

PR #205, opened 2015-09-13
Fix for column name not formed correctly when namespaced when constructing rows

robskillington:


stephencelis:
Thanks for the PR! It totally flew past my radar and I already got the column name fix in earlier, sorry! :disappointed:

As for the "embedded content contains swift" setting, can you provide more information?

robskillington:
@stephencelis the embedded swift content flag is due to app store submission requiring the flag because of the `sqlcipher` framework that is required and bundled.

PR #237, opened 2015-09-29
fixes #235

### LINKED ISSUE #235: Connection::run(query: Insert) failed to unwrap lastInsertRowid with OnConflict.Ignore
branch: swift-2
commit: 8bd27bcd5d0e6461620e51a651156d57c1c2c6cb

Model

``` swift
class News {

    let id: Int
    let header: String

    init(row: Row) {
        id = row[NewsColumn.Id]
        header = row[NewsColumn.Header]
    }
}
```

Insert code

``` swift
try db.run(tableNews.insert(or: OnConflict.Ignore,
    NewsColumn.Id <- (rawNews["id"] as! Int),
    NewsColumn.Header <- (rawNews["name"] as! String),
))
```

This insert code will fail [here](https://github.com/stephencelis/SQLite.swift/blob/swift-2/Source/Typed/Query.swift#L951) if model with the same Id already exists.

asfdfdfd:


stephencelis:
Thanks for the pull request!

This function should really be throwing on failure to insert, not returning an optional. If It's not throwing then we'll need to figure out why. If you have the chance to investigate, please do! Otherwise I'll try to get to this soon. Thanks again!

asfdfdfd:
It is throwing, yes. But my query is not "INSERT" but "INSERT OR IGNORE". I've thought that in that case library should ignore collision. Am i wrong?

stephencelis:
@asfdfdfd Yeah. I unfortunately didn't think through that part of the API, so thanks for bringing it to my attention. Perhaps we should add a new method that returns an optional for IGNORE statements only.

stephencelis:
Gah, GitHub automatically closes PRs that are trying to merge against deleted branches (swift-2 was merged in).

PR #243, opened 2015-10-05
Fetch a result row as an associative array

abiaad:
Use case : 
```swift
let rows = db.prepare("SELECT * FROM table")
while (rows.next() != nil) {
   let value = rows["column_name"] as? String
}
```


abiaad:
Wrong branch.

Issue #212, opened 2015-09-18
Read row by name


abiaad:
I'm sorry I didn't find the information through the repo. My question is very simple.
I can't write this at this time :
```swift
let rows = db.prepare("SELECT * FROM table")
for row in rows {
   print(row["id"])
}
```
Is there a way to achieve this ?
Thx !

stephencelis:
Try this workaround:

```
let rows = db.prepare("SELECT * FROM table")
for row in rows {
   print(row[rows.columnNames.indexOf("id")!])
}
```

abiaad:
Cool !
Thx

PR #244, opened 2015-10-05
Fetch a result row as an associative array

abiaad:
Use case : 
```swift
let rows = db.prepare("SELECT * FROM table")
while (rows.next() != nil) {
   let value = rows["column_name"] as? String
}
```

### DIFF
--- a/SQLite/Statement.swift
+++ b/SQLite/Statement.swift
@@ -47,6 +47,12 @@ public final class Statement {
     public lazy var columnNames: [String] = (0..<Int32(self.columnCount)).map {
         String.fromCString(sqlite3_column_name(self.handle, $0))!
     }
+    
+    /// Handle associative getter on the current row
+    public subscript(columnName: String) -> Binding? {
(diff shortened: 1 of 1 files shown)


stephencelis:
Gah, GitHub automatically closes PRs that are trying to merge against deleted branches (swift-2 was merged in).

stephencelis:
Hm, could you please provide more information as to what you're trying to solve here? What's wrong with the following?

Is there a reason you need to call `Statement.next`? Why didn't the following work for you?

``` swift
let stmt = db.prepare("SELECT * FROM table")
for row in stmt {
    let value = row["column_name"] as? String
    // …
}
```

If you're worried about the extra work creating an array each `for`…`in`, you can access via the `row` cursor as such:

```
let stmt = db.prepare("SELECT * FROM table")
while try stmt.step() {
    let value = stmt.row["column_name"] as String
    // …
}
```

abiaad:
Ahah ok I didn't realize I can do it this way ...
And you didn't answer me that last time ;)  https://github.com/stephencelis/SQLite.swift/issues/212

PR #251, opened 2015-10-09
Solve [!] Invalid `SQLite.swift.podspec` file: cannot infer basepath

ferranvila:
The **require_relative** is not working, changed to **require** and setting the $LOAD_PATH previously


stephencelis:
Thank you! :fireworks:

PR #284, opened 2015-11-24
Add threading info to documentation

patr1ck:
Just some suggested changes. I was hitting some `database is locked` errors myself and it took a while to figure out what was up based on the current documentation. Figured I'd add some extra information about `busyTimeout` and `busyHandler` here.

### DIFF
--- a/Documentation/Index.md
+++ b/Documentation/Index.md
@@ -10,7 +10,7 @@
       - [Read-Write Databases](#read-write-databases)
       - [Read-Only Databases](#read-only-databases)
       - [In-Memory Databases](#in-memory-databases)
-    - [A Note on Thread-Safety](#a-note-on-thread-safety)
+      - [Thread-Safety](#thread-safety)
   - [Building Type-Safe SQL](#building-type-safe-sql)
     - [Expressions](#expressions)
       - [Compound Expressions](#compound-expressions)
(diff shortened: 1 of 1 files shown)


stephencelis:
Thanks for the pull!

This is actually a great topic for discussion Maybe the smarter default is to have a handler? I see the appeal of the library working in a vanilla state matching the core beneath it, but other libs like FMDB already set a default handler, and it's easy enough to nil out or override. What do you think?

patr1ck:
I would love to see a default handler! I don't have any experience with FMDB (only Core Data) so I can't really speak to what expectations people coming from that might have, but seeing as SQLite.swift is a higher level library, I don't think that sensible defaults would be a problem.

I think part of the problem I had was that the `database is locked` error isn't super helpful. The [official docs](https://www.sqlite.org/lockingv3.html) don't make any mention of the `sqlite3_busy_timeout()` or `sqlite3_busy_handler()` calls, nor does the `sync<T>()` method of Connection, so it wasn't until I explored that class a little more fully until I realized I even had the option.

That said, a default which points out that these properties exist and can be used would likely go a long way in helping others resolve similar issues.

patr1ck:
Oh and thanks for the quick merge! :+1:

PR #221, opened 2015-09-21
propagate errors in Statement initialization

kujenga:
Propagates up errors by replacing a `try!` in the init method for Statements with try and adding the necessary acknowledgements of this throughout the code.

see: https://github.com/stephencelis/SQLite.swift/issues/220

Issue #220, opened 2015-09-21
Swift 2: runtime error during failed statement preparation


kujenga:
I'm trying to access a db that is very occasionally locked by another process and am experiencing a runtime error at the following line: https://github.com/stephencelis/SQLite.swift/blob/swift-2/Source/Core/Statement.swift#L34

A simple way to recreate this issue is to open the SQLite db that your application will be reading using the command line `sqlite3` tool and place an exclusive lock on it using `PRAGMA locking_mode = EXCLUSIVE; BEGIN EXCLUSIVE;`. The run any code or test that also accesses the db.

Since statement preparation call can fail, it seems like the `try!` should instead be a `try`, allowing the error to propagate up and be handled appropriately. I've implemented this change in a [fork](https://github.com/kujenga/SQLite.swift). Happy to submit a pull request if you think that would be appropriate.

stephencelis:
That's fair. Have you tried setting `busyTimeout` or `busyHandler` as a workaround? https://github.com/stephencelis/SQLite.swift/blob/swift-2/Source/Core/Connection.swift#L345-L373

kujenga:
Yeah, I actually went there first but it didn't seem to have much of an effect. I'm no SQLite expert though, it's possible I'm doing something wrong. Here's a pared down version of the code in question.

```swift
let db = try Connection(dbURL.path!, readonly: true)

db.busyHandler({ tries in
    tries < 3
})

for row in db.prepare("SELECT example FROM table WHERE something == 'value'" { // runtime error here
    // processing
}
```

mikemee:
@kujenga Are you ok if this is closed, now that https://github.com/stephencelis/SQLite.swift/pull/290 has been merged?

kujenga:
@mikemee Yep, sounds good to me!

Issue #187, opened 2015-08-26
sqlite3_prepare errors are fatal


zmeyc:
In Statement's constructor try! statement makes impossible handling schema mismatch errors gracefully:

```swift
public final class Statement {
    init(_ connection: Connection, _ SQL: String) {
        self.connection = connection
        try! check(sqlite3_prepare_v2(connection.handle, SQL, -1, &handle, nil))
    }
```

This leads to crashes in some unexpected places such as:

```swift
do {
        let sql = table
            .filter(keyColumn == value)
            .update(setters)
        let rowsUpdated = try db.run(sql) // crash
} catch {
}
```

I propose considering making Statement's constructor throwable and replacing try! with try. This will lead to API changes in other places (pluck and some other functions will become throw-able), but will allow recovering from errors.

stephencelis:
@zmeyc Could you elaborate on the schema mismatch issue you're coming across? Do you not have control over the database file you're working with?

zmeyc:
@stephencelis I'm using the library to implement a backend. One of the tables didn't have a required column due to migration problems. When users invoke endpoints which try to update that table the server crashes. Ideally, it should handle this in a controllable manner, for example log the error and return "Internal server error". I'll probably try to fix this in a cloned repo to see how it impacts the rest of the library API.

stephencelis:
I've tried to make read-only queries throw fatal errors since they usually indicate a programming issue that can be solved without error handling. Is this something that could be addressed at the migration layer?

zmeyc:
I don’t think this can be solved at the migration layer. I understand that in most cases sqlite3_prepare failing indicates a programming error or hdd failing or db admin mistake etc. But I’m trying to avoid shutting down the entire app if the error is recoverable (in my case affects only one endpoint).

Another problem is that some functions are declared as throwable and user would expect to handle errors by declaring a do{}catch{} yet they shut down the app in some cases.

Imho, it’s better to propagate errors from underlying lib where possible, possibly adding helper methods with try!.
I.e. non-throwing “scalar()” and “checkedScalar() throws”. In this case the API can be revisited so that functions with “throws” actually return all errors. It can probably even remain compatible with the current API.

I’ve replaced “try!” with “try” in Statement’s constructor and it went smoother than I thought:
db.prepare and db.scalar became throwable, other changes were internal to the lib.

Looks like try! is used only in a few places:
  Statement’s constructor
  createFunction()
  createCollation()
  scalar()
  next()
[...]

PR #290, opened 2015-11-30
propagate errors in Statement initialization

kujenga:
Reopened against master, replacing https://github.com/stephencelis/SQLite.swift/pull/221

Propagates up errors by replacing a `try!` in the init method for Statements with try and adding the necessary acknowledgements of this throughout the code. See issue https://github.com/stephencelis/SQLite.swift/issues/220

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,15 +56,15 @@ let insert = users.insert(name <- "Alice", email <- "alice@mac.com")
 let rowid = try db.run(insert)
 // INSERT INTO "users" ("name", "email") VALUES ('Alice', 'alice@mac.com')
 
-for user in db.prepare(users) {
-    println("id: \(user[id]), name: \(user[name]), email: \(user[email])")
+for user in try db.prepare(users) {
+    print("id: \(user[id]), name: \(user[name]), email: \(user[email])")
     // id: 1, name: Optional("Alice"), email: alice@mac.com
 }
 // SELECT * FROM "users"
 
 let alice = users.filter(id == rowid)
 
-try db.run(alice.update(email <- email.replace("mac.com", "me.com")))
+try db.run(alice.update(email <- email.replace("mac.com", with: "me.com")))
 // UPDATE "users" SET "email" = replace("email", 'mac.com', 'me.com')
 // WHERE ("id" = 1)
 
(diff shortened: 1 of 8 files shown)


ryanholden8:
Any updates as to wether this will be merged into the repo? 
This seems to address #187, which is a similar situation I also ran into. 

It seems odd that a library would decide when my app should crash when really the context of the application code is needed to make that determination. (Just my viewpoint though) Thank you!

kujenga:
@ryanholden8 I just rebased my branch off the tip of master, so if you'd like you're welcome to use it in the interim. I won't be changing anything else on it.

@stephencelis What are your thoughts on pulling this in? Since it does break the API by it's nature, to me it seems the sooner the better.

stephencelis:
@kujenga Let's figure out the balance here. The reason I've been holding off is as follows:

1. On the one hand, recovering from errors is important when you're dealing with validation and other recoverable failures.
2. On the other hand, SQL statements that read from the database generally shouldn't fail unless the statement is invalid (_e.g._, a syntax error, which is a developer issue).

Because of this, I'm wondering if we should only promote `prepare` to be `throws`, but let the others remain fatal. With this change, developers that need the flexibility of error recovery can use `db.prepare`, but the rest can generally use the simpler, non-throwing APIs (`scalar`, `pluck`, etc.). What do you think?

kujenga:
@stephencelis Sounds good! I'm in agreement that developer error doesn't necessitate runtime error handling. The `scalar` method is easy to keep as is, since I converted that one over [separately](https://github.com/kujenga/SQLite.swift/commit/ef187a6d5d55315834c05823a266b0ad5ad4fd4f). `pluck` does call `prepare`, but I can just add a `try!` to the guts of that method to provide the proper API. I'll also put a short note in the comments for prepare indicating the reason that it throws an error while the others don't.

Coming right up! 😃

stephencelis:
Great! Let me know when everything's pushed and I'll take a final look and merge :+1:

kujenga:
@stephencelis Alright! After some liberal rebasing this should be cut down to a more manageable set of changes.

stephencelis:
Looks good! Thanks for this! :100:

PR #304, opened 2015-12-25
correct SQLs in the doc's Setters section

liushigit:


stephencelis:
Thanks for the pull request! I guess some of the doc examples are confusing without greater context, but this example wasn't actually a typo :frowning:

In this case, there's a `counters` table and `counter` is just a single row. Here's an example of greater context that should run in a playground (assuming you wrap it in a `do { … }` block):

``` swift
let db = try Database()

// columns
let id = Expression<Int64>("id")
let count = Expression<Int>("count")

// table
let counters = Table("counters")

// create table
try db.run(counters.createTable { t in
    t.column(id, primaryKey: true)
    t.column(count, default: 0)
})

// insert row and assign filtered query
let counter = counters.filter(id == try db.run(counters.insert()))

// example
try db.run(counter.update(count++))
```

Make sense? Do you think providing additional context would improve the docs? Was there something specific you found confusing?

stephencelis:
One thing to note: I'll probably be removing the `++` and `--` functions (since Swift core is removing the operators in Swift 3).

stephencelis:
Closing this out for now. I'd still love feedback and would be happy to reopen this if you have a clearer way of communicating the context! Thanks again for the pull.

PR #307, opened 2016-01-05
Add order method to query that takes an array arg

cjwirth:
The same thing exists for `group` so I figured not having one for `order` might have been an oversight.

Copy/pasted the documentation from above (just like `group`).
Tests still pass.

:smile:

### DIFF
--- a/Source/Typed/Query.swift
+++ b/Source/Typed/Query.swift
@@ -399,6 +399,22 @@ extension QueryType {
     ///
     /// - Returns: A query with the given `ORDER BY` clause applied.
     public func order(by: Expressible...) -> Self {
+        return order(by)
+    }
+    
+    /// Sets an `ORDER BY` clause on the query.
+    ///
+    ///     let users = Table("users")
+    ///     let email = Expression<String>("email")
+    ///     let name = Expression<String?>("name")
+    ///
+    ///     users.order([email.desc, name.asc])
+    ///     // SELECT * FROM "users" ORDER BY "email" DESC, "name" ASC
+    ///
+    /// - Parameter by: An ordered list of columns and directions to sort by.
+    ///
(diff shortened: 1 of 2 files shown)


Source/Typed/Query.swift
```diff
@@ -399,6 +399,22 @@ extension QueryType {
...
     /// - Returns: A query with the given `ORDER BY` clause applied.
     public func order(by: Expressible...) -> Self {
+        return order(by)
+    }
+    
+    /// Sets an `ORDER BY` clause on the query.
+    ///
+    ///     let users = Table("users")
+    ///     let email = Expression<String>("email")
+    ///     let email = Expression<String?>("name")
```
stephencelis: Should this be `let name =`?

Source/Typed/Query.swift
```diff
@@ -399,6 +399,22 @@ extension QueryType {
...
+    ///
+    ///     users.order(email.desc, name.asc)
```
stephencelis: Should this demonstrate use with an array?

stephencelis:
Thanks! I really wish Swift let you splat out variadic args, but till that's supported, this is needed!

Could you fixup/rebase the documentation and, if possible, add a test that calls `order` with an array? Thanks!

cjwirth:
Ah, good eye, sorry about that!
Fixed up the documentation and added a test for it.

Source/Typed/Query.swift
```diff
@@ -390,7 +390,7 @@ extension QueryType {
     ///
     ///     let users = Table("users")
     ///     let email = Expression<String>("email")
-    ///     let email = Expression<String?>("name")
+    ///     let name = Expression<String?>("name")
```
stephencelis: :sweat_smile: Thanks!

stephencelis:
Thanks for your help!

PR #309, opened 2016-01-06
Add Carthage install info to main documentation

mikemee:
Copied the Carthage install from the readme for people who end up here directly.  _Not terribly DRY, but short of trimming the front page or linking back to it, it's arguably an improvement._


stephencelis:
Already wasn't DRY! Thanks for fixing this omission! :+1:

Issue #311, opened 2016-01-06
Move SQLCipher support into separate repo


mikemee:
Per discussion on #192, I'm opening this as a catch-all specific issue to track splitting out the [SQLCipher](https://github.com/sqlcipher/sqlcipher) support into a separate repo.

> I need to extract SQLCipher support out into its own repo. It's causing too many problems when bundled.

This appears to affects several Carthage and Cocoa Pod build issues (e.g. #300, #308) as well as dependent features (#272), many of which could at least be isolated to just those who need `SQLCipher`.

mikemee:
Completed via PR #314. 

Replacement SQLCipher available for testing at https://github.com/stephencelis/SQLiteCipher.swift. Comments and PR's welcomed!

Issue #272, opened 2015-11-16
Need TV OS Support


srkrambo:
I get following  errors when trying to build for TV os 

 "Target 'SQLiteCipher.swift' of project 'Pods' was rejected as an implicit dependency for 'SQLiteCipher.framework' because it doesn't contain platform 'appletvsimulator' in its SUPPORTED_PLATFORMS 'iphonesimulator, iphoneos'"

There is a similar error for "Target 'SQLCipher' of project 'Pods' was rejected as an implicit dependency for 'SQLCipher.framework' because it doesn't contain platform 'appletvsimulator' in its SUPPORTED_PLATFORMS 'iphonesimulator, iphoneos'".

Here in Cocoapods, http://blog.cocoapods.org/CocoaPods-0.39/
Its mentioned we need to explicitly add in podspec to support TV OS.

stephencelis:
I have a fix for this in the works, but it looks like the upstream version of SQLCipher hasn't had its project file updated for tvOS or watchOS support.

mikemee:
Update: [SQLCipher](https://github.com/sqlcipher/sqlcipher) has merged support into their `prerelease` branch on Nov 17 (here 71691b6a1ccd2a7ae2d0fa8f03ad822ae6aaac9e), but they haven't merged to `master` since July :(

srkrambo:
Yes @mikemee, I am keeping an eye on the release of SQLCipher, will reach out to again once they release next version.

stephencelis:
Maybe we can point the submodule to `prerelease` in the meantime?

mikemee:
@srkrambo Do you need the SQLCipher support for your tvOS project (i.e. encrypted data in your sqlite database)?

If not, please try the branch in PR https://github.com/stephencelis/SQLite.swift/pull/314 that I just submitted and please let me know if that works for you. It should be trivial to add a tvos target. 

It would be nice to confirm it works and I'll add the target into the build if the PR is accepted and once the dust settles.

rajkumariosdev:
@mikemee, we do need SQLCipher support since it contains data if exposed can be used by competitor apps.

srkrambo:
@mikemee, we do need SQLCipher support since it contains data if exposed can be used by competitor apps.

mikemee:
Ah, fair enough.  Not trusting the sandbox eh :)  (Or doesn't it exist on tv?)

[...]

Issue #230, opened 2015-09-23
Umbrella header SQLite.h not found in module.modulemap


arnoldschmid:
I've updated XCode to the new version and I am trying to convert everything to use Swift 2.0. Unfortunately SQLite does not work anymore. I used the newest cocoapods (ver. 0.38.2) and i use this podfile:
use_frameworks!
pod 'SQLite.swift',git: 'https://github.com/stephencelis/SQLite.swift.git',branch: 'swift-2'
But I receive always the above mentioned error. I tried to clean the Build Folder without success, always the same. Using the old version of SQLite.swift (copying the project in my project) with the old version of xcode was running like a charm. Now I have troubles using SQLite.swift. 
thanks
arnold

vineethvijayan:
@arnoldschmid  do u have the below issue

header "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/sqlite3.h"

??
I am also facing the same

hohband:
Does anybody solve this problem? I have the same problem.

stephencelis:
Can you try against the master branch again? I've merged the swift-2 branch in.

warn4n:
We are still having this issue on the master branch. We had earlier failed to get it working with pods on Swift 1.2, so had fallen back to the manual installation instructions. We are really hoping to go the pods route now that we are upgrading our project to Swift 2. We think we thoroughly removed the previous manual installation, but I guess there is a chance this issue could be related to some ghost or remnant of that.

stephencelis:
@warn4n Can you tell me where your Xcode.app is installed?

warn4n:
It's in /Applications, updated from the Mac app store.

coolgastjen:
I can confirm this problem. It happened after updating to Swift 2.
Xcode.App is also in the /Applications folder

cliftonlabrum:
Did anyone figure this out? I just installed this today in Xcode 7.1.1 and I'm getting this same error.

nickmshelley:
@cliftonlabrum I'm pretty sure the swift2 branch was merged into master and is out of date. We have this line in our podfile and it works fine: `pod 'SQLite.swift', git: 'https://github.com/stephencelis/SQLite.swift.git'`

cliftonlabrum:
That's what I have in my Podfile, too. Maybe something is wrong with my build settings in Xcode.

[...]

Issue #202, opened 2015-09-11
Build fails when the project path contains whitespaces


sebastianvarela:
Hi!
If I clone the repository and my path contains a whitespace on some directory name, I can't compile the  SQLiteCipher iOS Schema. I tried removing the space that I had and the project then builds correctly.
This is important because Jenkins directories usually contain spaces.

I'm building with Xcode GM using the following commit in the swift-2 branch:
Commit 4d5f4a11d20dc02bbf65aa2a4a8cf6c8d616ee12

Log:
```
The following build commands failed:
	CompileC /Users/Shared/Jenkins/Library/Developer/Xcode/DerivedData/SQLite-ggbbmcwgetpvjxgnjfngxdjdnqex/Build/Intermediates/sqlcipher.build/Release-iphonesimulator/sqlcipher.build/Objects-normal/x86_64/sqlite3.o sqlite3.c normal x86_64 c com.apple.compilers.llvm.clang.1_0.compiler
	CompileC /Users/Shared/Jenkins/Library/Developer/Xcode/DerivedData/SQLite-ggbbmcwgetpvjxgnjfngxdjdnqex/Build/Intermediates/sqlcipher.build/Release-iphonesimulator/sqlcipher.build/Objects-normal/i386/sqlite3.o sqlite3.c normal i386 c com.apple.compilers.llvm.clang.1_0.compiler
(2 failures)
ld: warning: directory not found for option '-F/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator9.0.sdk/Developer/Library/Frameworks'
ld: warning: directory not found for option '-F/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator9.0.sdk/Developer/Library/Frameworks'
ld: warning: directory not found for option '-F/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS9.0.sdk/Developer/Library/Frameworks'
ld: warning: directory not found for option '-F/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS9.0.sdk/Developer/Library/Frameworks'
[...]

PR #314, opened 2016-01-08
Prep for public release to CocoaPods

mikemee:
@stephencelis, as discussed, here is a branch with a proposed update to become the first "public" cocoapod version.  

### Summary of changes:

 * Remove `SQLiteCipher` (for now, per #311)
 * rebuilt clean from "New Project" with Xcode 7.2 (_it removed some cruft required by older Xcode releases & hopefully also fixes some lingering build issues_)
 * fix build glitches with Carthage (e.g. #202)
 * fix build glitches with CocoaPods (e.g. #230)
 * separate build targets from test targets with @testable import (added in Swift 2.0)
 * make sqlite3 a Swift system module so `import SQLite3` works in `.swift` files (looking ahead, this is [how the swift package manager will do it](https://github.com/apple/swift-package-manager/blob/master/Documentation/SystemModules.md))
 * becomes an Xcode Project instead of Workspace
 * folder structure rearranged to correspond to an Xcode "New Project" layout

### Known CocoaPod hassles:
 * I couldn't convince it to generate its own `module.modulemap` file, so one is included
 * has spurious warnings about a missing include: `Umbrella header for module 'SQLite' does not include header 'fts3_tokenizer.h` (all tests, including tokenizer, run fine)
 * `pod lib lint` does not work to validate the pod, but `pod spec lint` is fine (per https://github.com/CocoaPods/CocoaPods/issues/4607) (Note: does not currently pass either as it needs a branch, but it did pass in testing on my repo with the url modified).

### Next Steps
If this PR is accepted, the following should be done to make CocoaPods and Carthage happy:
 * tag the commit in `master` with '0.9.0'
 * push to CocoaPods per: https://guides.cocoapods.org/making/getting-setup-with-trunk.html

### Trying this PR now

I encourage others to try this PR. Because it's on a branch, it is a trivial change to your existing files (or manually as before):

Carthage `Cartfile`:
```
github "stephencelis/SQLite.swift" "publicpod"
```
CocoaPod `Podfile`:
```
use_frameworks!

pod 'SQLite.swift',
  git: 'https://github.com/stephencelis/SQLite.swift.git', :branch => 'publicpod'
```
### After this
 * The cleaned up build should make it trivial to add TV support (#272) (_very tempted to just add it now..._)
 * Create a new repo that builds on this to restore SQLiteCipher via https://github.com/sqlcipher/sqlcipher

### Other TODOs

- [ ] Remove SQLCipher references and installation instructions from README, Documentation


.gitignore
```diff
@@ -1,10 +1,23 @@
+# OS X
+.DS_Store
```
stephencelis: Good change. Prepares us for swiftpm compatibility :+1:

Makefile
```diff
@@ -48,6 +48,6 @@ repl:
 		swift -F '$(TMPDIR)/SQLite.swift/Build/Products/Debug'
 
 sloc:
-	@zsh -c "grep -vE '^ *//|^$$' SQLite/*.{swift,h,c} | wc -l"
+	@zsh -c "grep -vE '^ *//|^$$' SQLite/*/*.{swift,h,m} | wc -l"
```
stephencelis: Good catch! I don't think the depth is deeper than one directory here at the moment, but maybe `**/*` for future safety?

mikemee: I tried "make sloc" and only got errors, and this made it work.  (_I almost deleted everything from the Makefile to "start fresh", but didn't want to deal with the potential repercussions as I'd already changed so much!  (cough)bike shedding(/cough)_  )

.gitignore
mikemee: Thx :) I can't take full credit. This just seemed to be the norm in similar projects I viewed for comparison when I did "new project", so I copy pasted. Plus I've removed more than a few `.DS_Store` files over time.

SQLite.swift.podspec
```diff
@@ -1,17 +1,32 @@
-$LOAD_PATH << '.'
-require 'Supporting Files/podspec.rb'
+#
+# `pod lib lint SQLite.swift.podspec' fails - see
```
stephencelis: Hm, I'm getting this:

```
 -> SQLite.swift (0.9.0)
    - ERROR | xcodebuild: Returned an unsuccessful exit code.
    - ERROR | [iOS] swift: Swift support uses dynamic frameworks and is therefore only supported on iOS > 8.
    - ERROR | xcodebuild:  /Users/stephencelis/Developer/SQLite.swift/SQLite/Core/Blob.swift:25:8: error: no such module 'SQLite3'

[!] SQLite.swift did not pass validation, due to 3 errors.
```

I can eliminate the middle one by adding this line:

``` rb
  s.ios.deployment_target = "8.0"
```

The other two, when running `--verbose`, are failing on the `import SQLite3`, assuming because there's no `0.9.0` tag with the `SQLite3` directory/sub-pod available?

SQLite3/module.modulemap
```diff
@@ -0,0 +1,5 @@
+module SQLite3 [system] {
+    header "/usr/include/sqlite3.h"
```
stephencelis: I think it's safer to have a module.modulemap per platform, though this can become problematic when dealing with Xcode-beta.app. I know the SwiftPM team is attempting to solve the static module map issue in the future.

```
  s.ios.module_map = "Supporting Files/Module Maps/ios/module.modulemap"
  s.osx.module_map = "Supporting Files/Module Maps/osx/module.modulemap"
```

SQLiteTests/AggregateFunctionsTests.swift
```diff
@@ -1,5 +1,5 @@
 import XCTest
-import SQLite
+@testable import SQLite
```
stephencelis: I generally avoid `@testable` in frameworks till I need it. I try to test the public API only. Were you encountering an issue running tests without this attribute?

module.modulemap
```diff
@@ -0,0 +1,12 @@
+// This file is required by current CocoaPods (0.39.0). 
+//  - Generates warning: Umbrella header for module 'SQLite' does not include header 'fts3_tokenizer.h'
```
stephencelis: I wonder if we can make this a private header in the module to avoid the warning?

stephencelis:
This is great! Thanks for all the work you put into it! I left a few comments/questions, but I think this is really close to being ready.

SQLite.swift.podspec
mikemee: Are you doing "pod lib lint ..." perhaps?  You need to use "pod spec lint ...", per https://github.com/svdo/swift-netutils/issues/4 which leads to https://github.com/CocoaPods/CocoaPods/issues/4607. See comment on line 2. That same error cost me several hours until I found the pre-existing issue :(

I agree that the deployment target is a good idea regardless though. (Though to be honest, I'm not sure it's possible to have Swift on < 8.0 anyway, so I think this comes out in the wash).

SQLite3/module.modulemap
mikemee: Not against this. Was just hoping to avoid a solution like this that runs a script to adjust the path if needed: 

https://github.com/svdo/swift-netutils/tree/master/ifaddrs

Unless you can think of a better way?

SQLiteTests/AggregateFunctionsTests.swift
mikemee: Good point.  I was doing this in another project so I just did the same here. Likely not needed at all.  Note, FYI, that I did change from including SQLite as part of the test target in Xcode, which is was previously.

I'll try without and push an update if it works ok, or add a comment here if not.

module.modulemap
mikemee: I tried all sorts of things. CocoaPods is a magical mess was the only conclusion I could come up with. I essentially had to choose between keeping Xcode happy, keeping CocoaPods happy, or crawling under the bed to cry. I hoped that once this PR went "live", someone would point out the stupidity of my ways and offer a fix!

That said, I've somewhat recovered from the trauma, and will see if making it private makes the problem go away! ;)

SQLite3/module.modulemap
stephencelis: A script would be great. Is that possible given CocoaPods infrastructure?

SQLiteTests/AggregateFunctionsTests.swift
stephencelis: :+1:

module.modulemap
stephencelis: In the spirit of progress, let's merge first and ask questions...soon :)

SQLite3/module.modulemap
mikemee: According to the svdo/swift-netutils link, yes, this can be made to work just for CocoaPods. I'll give it a try.

stephencelis:
This'll need to get rebased (given #317). Once you think things are ready, give me a nod, I'll take a final look and merge!

SQLiteTests/AggregateFunctionsTests.swift
mikemee: Works just fine. Prepping an update.

stephencelis:
(Oh, and as far as a separate SQLiteCipher.swift project goes, should I create a repo and add you?)

mikemee:
Rebase is done. @testable done. deployment target added to podspec. 

Working on the other items.  (Trying private for fts3_tokenizer.h just to see... so far Xcode happy, but still need to test Carthage (likely fine since Xcode is) and of course CocoaPods to be sure).

Re SQLiteCipher - sure. Can't promise anything, but I do like to see things through... :)

stephencelis:
Yeah, I think the other repo at least needs to work very basically since people depend on SQLCipher in their projects. Let me see if I can get something basic working against this branch.

I also just noticed that SQLCipher installation instructions are still in the README/Documentation on this branch; we should remove/update them before merging.

mikemee:
Private for `fts3_tokenizer.h` still generates the same warning, so I'll leave it as is (`project`), unless you have strong feelings about it.

Will fix the readme/documentation re SQLCipher

Re the module map. I just realized that even if I fix it via a magic script for Cocoa, it's still "wrong" for Swift Package Manager / current Xcode style file in `SQLite3/module.modulemap`, i.e.:
```
module SQLite3 [system] {
    header "/usr/include/sqlite3.h"
    link "sqlite3"
    export *
}
```
And I don't know any simple way of changing that, hmm, except, I suppose by setting up multiple files, then changing the Swift Include file path in Xcode... all of which seems more messy for a problem that will go away with Swift Package Manager soon enough anyway (I hope!)

I.e. these are all identical on my machine:
```
/usr/include/sqlite3.h 
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/usr/include/sqlite3.h
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/sqlite3.h 
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk/usr/include/sqlite3.h
```
Given that all these files are identical anyway, I propose sticking with the original checkin and not complicating the CocoaPod build further.  

But I'll accept whatever you prefer ... it's your project (_though I'm not excited to hack the script I will admit!_ ;)

stephencelis:
:100:

SQLite3/module.modulemap
iNoles: all it do "prepare_command": "ifaddrs/injectXcodePath.sh"

mikemee: Yes, thanks! See my comment in the PR 314 thread, starting with:
> Re the module map. I just realized that ....

PR #329, opened 2016-01-14
Add support links to features, add link to SQLiteCipher, Gitter link

### LINKED ISSUE #328: Idea: Add gitter chat room
It appears that https://gitter.im/ provides a free chat room that you can associate with an open source Github project. I've seen it in a of couple projects I use.  Not sure if this project has enough active users to make it worthwhile, but it might be interesting to try.  I'm willing to keep a window open sometimes to see how it goes.

@stephencelis I think you need to do the signup, after which we can add a badge to the readme (I suggest right near the top).

_Of course, one might argue that the last thing anyone needs is another chat room, but it might be useful and costs little effort to try... we can always close it if it turns out to be unproductive._

mikemee:
Fixes #328


stephencelis:
:100:

Issue #332, opened 2016-01-19
expression unexpectedly raised an error: no such table


dericeira:
Hello.

Follow my code:

   let path = NSSearchPathForDirectoriesInDomains(
            .DocumentDirectory, .UserDomainMask, true
            ).first!
        
        let db = try! Connection("\(path)/main.sqlite3")

        db.trace { print($0) }
        
        let words = Table("words")
        let word = Expression<String>("word")
        
        
        for user in try! db.prepare(words) {
            print("id: \(user[word])")
        }

Returning this error:

fatal error: 'try!' expression unexpectedly raised an error: no such table: words: file /Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-700.1.101.15/src/swift/stdlib/public/core/ErrorType.swift, line 50

but IF I use a different way to connect (read only) I have success:

let path = NSBundle.mainBundle().pathForResource("main", ofType: "sqlite3")!
let db = try Connection(path, readonly: true)

BUT I need it read and write. 

What I am doing wrong?

mikemee:
Is this for iOS or OSX (or TV)?  

As described in this documentation, to have a writeable database, iOS requires you to create the database in the Documents directory.  Anything you ship in a bundle with your app will be read-only. 

See https://github.com/stephencelis/SQLite.swift/blob/master/Documentation/Index.md#connecting-to-a-database for information.

Is this your scenario?  Or do you think there is still a framework bug here somewhere?  Thanks!

dericeira:
Yeah, iOS.

But I'm using this example:

mikemee:
Your code needs to copy the database to the Documents folder. Your app needs to do that if you are going to modify the database.  See these two links for details:

https://stackoverflow.com/questions/34609746/what-different-between-store-database-in-different-locations-in-ios

https://stackoverflow.com/questions/34614968/ios-how-to-copy-pre-seeded-database-at-the-first-running-app-with-sqlite-swift

Once you have this working, we'd love to have some code to add to the documentation to help the next person with this question!

ihula:
Your code needs to copy the database to the Documents folder. 
![2016-01-20 09 43 41](https://cloud.githubusercontent.com/assets/11713485/12437644/8c18bf5e-bf5a-11e5-80bb-1567da93bf49.png)


let path = NSBundle.mainBundle().pathForResource("Documents/Light.sqlite", ofType: nil)!
self.db = try! Connection(path)

PR #333, opened 2016-01-20
Added links to SO articles with more iOS info

mikemee:
Based on the recent #332 issue, and earlier similar questions, I've added this note to the documentation. Linking to SO is a short-term measure, but if it stops new issues being opened, then everyone wins imnsho.


stephencelis:
I'm not entirely sure this will eliminate questions of this nature. But we can hope! :+1:

PR #336, opened 2016-01-21
Fixes #335 - allow OSX 10.9 and above instead of 10.11

### LINKED ISSUE #335: Carthage and OS X 10.10
I installed 0.9.2 via Carthage, get the following error:

```
.../MainWindowController.swift:11:8: Module file's minimum deployment target is OS X v10.11: .../Carthage/Build/Mac/SQLite.framework/Modules/SQLite.swiftmodule/x86_64.swiftmodule
```

Which means that it cannot run on OS X older than 10.11. I suggest to also support 10.10 (and probably 10.9).

mikemee:
Reminder: will need to tag this update for Carthage to be happy.


stephencelis:
:+1:

PR #344, opened 2016-02-02
Fix no Foundation import error

banxi1988:
withouth import Foundation sometimes will cause Compile Error


ryanholden8:
Also ran into this issue, and fixed it the same way. 

This does not happen when I use SQLite as a module. 
Happened when I imported the SQLite source files directly due to the fact that I still need to support iOS7. :(

stephencelis:
Thanks!

PR #352, opened 2016-02-07
Fix Carthage build errors on case-sensitive filesystem

mkistler:
SQLite fails to build with Carthage on Macs with case sensitive file system.  The build fails with:

    The following build commands failed:
    	ProcessInfoPlistFile /Users/mkistler/Library/Developer/Xcode/DerivedData/SQLite-fciegensodowgvfsgoskujfmqkpa/Build/Products/Release/SQLite.framework/Versions/A/Resources/Info.plist SQlite/Info.plist
    (1 failure)
    error: could not read data from '/Users/mkistler/Projects/myproj/Carthage/Checkouts/SQLite.swift/SQlite/Info.plist': The file “Info.plist” couldn’t be opened because there is no such file.
    A shell task failed with exit code 65:
    ** BUILD FAILED **

The nut of the problem is that the "L" in "SQLite" must be capitalized or the build will fail on Macs that have case-sensitive filesystems.

### DIFF
--- a/SQLite.xcodeproj/project.pbxproj
+++ b/SQLite.xcodeproj/project.pbxproj
@@ -100,7 +100,7 @@
 			containerPortal = EE247ACA1C3F04ED00AE3E12 /* Project object */;
 			proxyType = 1;
 			remoteGlobalIDString = EE247B3B1C3F3ED000AE3E12;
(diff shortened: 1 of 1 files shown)


stephencelis:
Nice catch, thanks!

PR #354, opened 2016-02-08
Added support for SQLITE_OPEN_SHAREDCACHE to facilitate migrating from other sqlite libraries

andrewjackman:
I have added a sharedCache flag to Connection initializers. This allows multiple connections to be made to the same database within a process. If true, the "SQLITE_OPEN_SHAREDCACHE" bitmask is ORed to the flags passed to the database open function call. This can be useful if gradually migrating from another sqlite3 library to SQLite, so that both libraries can open the same database and read/write to it.

See here for further information: https://sqlite.org/sharedcache.html

### DIFF
--- a/SQLite/Core/Connection.swift
+++ b/SQLite/Core/Connection.swift
@@ -65,10 +65,15 @@ public final class Connection {
     ///
     ///     Default: `false`.
     ///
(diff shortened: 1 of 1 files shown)


jberkel:
I think it would be better to expose all possible flags to `sqlite3_open_v2` as an OptionSet which gets passed into the initializer.

PR #357, opened 2016-02-13
Add target for tvOS

mkistler:
This PR adds build and test targets for tvOS.  This allows SQLite to be included in tvOS projects using carthage. All tests pass on the tvOS target.

I see that the podspec already claims tvOS support -- I didn't test this but it seems unlikely that these changes would break it.


stephencelis:
Thanks for another PR! I've merged #355 since it came in first, but the pull is still much appreciated!

PR #361, opened 2016-02-19
Made public the function bind(value:atIndex:) of class Statement

aaresu:
I have made public the function to use it when I want to research better performance for working with large dataset.


jberkel:
Why is this needed?

NathanFallet (approved):
Need to check if new CI tests are passing, but looks great!

jberkel:
As I said before, I'm not sure what this is trying to solve. There's also a pointless addition of a file already present in the repo (possibly due to merge).

NathanFallet:
So this should not be merged?

jberkel:
No, let's close this.

PR #375, opened 2016-03-19
Swift22 fixes

SebastianOsinski:
This PR resolves build errors in test target connected with change in declaration of one of XCTest methods (String was replaced with StaticString). It also resolves all warnings about deprecation of ++ operators etc. This PR should be merged after official release of Xcode 7.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ before_install:
   - gem install xcpretty --no-document
 script:
   - make test
-osx_image: xcode7
+osx_image: xcode7.3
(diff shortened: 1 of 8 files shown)


stephencelis:
Thank you! Does travis support Xcode 7.3? Wonder if bumping the `.travis.yml` file would allow tests to run?

SebastianOsinski:
I am not sure unfortunately. We should probably wait till official release.

stephencelis:
Thanks!

SebastianOsinski:
@stephencelis Maybe it would be good idea to mark current HEAD as 0.93 version?

stephencelis:
@SebastianOsinski I was hoping to resolve CocoaPods issues before bumping the version (see #380).

SebastianOsinski:
@stephencelis You're right, it sounds like a good idea, especially Carthage builds current version without a problem on Xcode 7.3.

PR #379, opened 2016-03-23
Update module.modulemap for Xcode 7.3

fuxx:
Imports the SDK based on current architecture. Fixes incompatibility problems for Xcode 7.3.
Also includes link flag for convenience.

### DIFF
--- a/podstuff/module.modulemap
+++ b/podstuff/module.modulemap
@@ -5,7 +5,18 @@ framework module SQLite {
     //
     //     header "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk/usr/include/sqlite3.h"
     //     header "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk/usr/include/sqlite3.h"
-    header "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/sqlite3.h"
+    //     header "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/sqlite3.h"
+    module arm {
+        header "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/sqlite3.h"
+        link "sqlite3"
+        requires arm
+    }
+    
+    module x86 {
+        header "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk/usr/include/sqlite3.h"
+        link "sqlite3"
+        requires x86
+    }
 
     export *
     module * { export * }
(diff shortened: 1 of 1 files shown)


fuxx:
Please test it in your environment. This works for my project which lives in a similar environment.

fuxx:
Seems not to be backward compatible with Xcode 7.2.1

fuxx:
I missed to add arm64, works now with Xcode 7.2.1 too. Backward compatibility should work back to Xcode 6.x

stephencelis:
@fuxx Nice solution! Looks like additional architectures may need to be added, though. Mac builds don't appear to work. Haven't had time to check tvOS/watchOS.

stephencelis:
tvOS seems to fail, too. I quickly tried adding `x86_64` and `i386` modules for Mac, but that doesn't seem to satisfy things.

fuxx:
@stephencelis i will try it out and i will make another commit. This was a quick shot from my results when i fixed the problem for my framework. i386 should be x86. 

i like the solution as it does not require to maintain multiple module maps as you could do in the cocoa pods specs.

stephencelis:
@fuxx Much appreciated! You solution is definitely more elegant than the multiple module map approach! Module maps in general are a brittle solution to this problem, though, so I hope CocoaPods adds support for better approaches in the future.

gblotter:
@stephencelis - When this PR get's approved and merged... what are the chances of it making it into your SQLCipher.swift repo?

stephencelis:
@gblotter The SQLCipher repo is unfortunately still a WIP. I'm happy to help contributors along to get it up and running (let me know if you're interested!), but won't have enough time to dedicate to release myself.

gblotter:
@stephencelis fair enough.  I think at the very least I'll submit a PR for a pod file... in hopes that that allows it to get traction with others using it.  Thanks for the response!

stephencelis:
@fuxx I was able to work around the issue with a combination of your solution and multiple module maps:

``` ruby
spec.ios.module_map = 'podstuff/ios.modulemap'
spec.osx.module_map = '…'
spec.tvos.module_map = '…'
```

Not ideal, but the cleanest approach I've found so far. I'll merge this in soon with the additional files unless you have a better solution!

fuxx:
@stephencelis unfortunately my solution will not work as standalone. As module maps do not allow branching, so i cant control the system platform. Even requiring the existence of different CPU instructions (like SSE or AES) per module wont work here, because the simulator on OS X would match that requirement for OS X too.

:( i will close the pull request, sorry ;(

fuxx:
Who ever is interested into possible restrictions in module maps could take a look here.
http://clang.llvm.org/doxygen/Basic_2Module_8cpp_source.html#l00060

Line 62. Target options are supported cpu instructions e.g. AES, SSE, SSE2, MMX etc.

stephencelis:
@fuxx I'll be merging in the beginning of your work soon! I've modified it so that CocoaPods should pick the right module map per platform. Please take a look at https://github.com/stephencelis/SQLite.swift/compare/cocoapods-xcode-7-3 and let me know if I'm missing anything!

fuxx:
@stephencelis 

You should remove line 5 (requires keyword) from osx module map. The rest looks fine for me.

https://github.com/stephencelis/SQLite.swift/compare/cocoapods-xcode-7-3#diff-c73d9f82bddf197c9492f3c7dbd63170R5

stephencelis:
Don't know how that slipped in! Thanks!

PR #381, opened 2016-03-23
Adds watchOS support

acrookston:
Based on the new Xcode 7.3 fix on cocoapods-xcode-7-3.

Regarding tests: the existing ones pass, but it seems like there is no test suite for watch os apps. If somebody knows how to add them, I'd be happy to do it, but I just can't find that option.
However I've successfully integrated this version in my watch app which used an older version of SQLite.swift.

Let me know if there's anything else missing, I'd be happy to amend.

PR #385, opened 2016-03-25
WatchOS support

acrookston:
Replacing and Closing #381 now based against master.

Regarding tests: the existing ones pass, but it seems like there is no test suite for watch os apps. If somebody knows how to add them, I'd be happy to do it, but I just can't find that option.
However I've successfully integrated this version in my watch app which used an older version of SQLite.swift.

Let me know if there's anything else missing, I'd be happy to amend.


stephencelis:
:bow:

PR #380, opened 2016-03-23
CocoaPods fix: Xcode 7.3

stephencelis:
This should hopefully fix issues #349 #378 and includes work from #379.

Please test on your platform and report if it works!

``` ruby
pod "SQLite.swift",
  git: "https://github.com/stephencelis/SQLite.swift",
  branch: "cocoapods-xcode-7-3"
```

Will merge when things look good!

Issue #349, opened 2016-02-05
Build failing with Xcode 7.3 beta


vhain:
My configuration is:
- Using latest Xcode 7.3 beta (7D129n)
- Using latest version of Cocoapods (0.39.0)
- with Podfile:
  ```
use_frameworks!

target 'app' do

  pod 'SQLite.swift'

end
```

I tried with or without `xcode-select -s [PATH_TO_XCODE_BETA]` or `xcode-select -s [PATH_TO_XCODE]`, but they both failed by following reason:
```
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/module.modulemap:1:8: error: redefinition of module 'Compression'
module Compression [system] [extern_c] {
       ^
/Applications/Xcode-beta.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator9.3.sdk/usr/include/module.modulemap:1:8: note: previously defined here
module Compression [system] [extern_c] {
       ^
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/module.modulemap:6:8: error: redefinition of module 'Darwin'
module Darwin [system] [extern_c] {
       ^
/Applications/Xcode-beta.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator9.3.sdk/usr/include/module.modulemap:6:8: note: previously defined here
module Darwin [system] [extern_c] {
       ^
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/module.modulemap:1478:8: error: redefinition of module 'os'
module os [system] [extern_c] {
       ^
/Applications/Xcode-beta.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator9.3.sdk/usr/include/module.modulemap:1599:8: note: previously defined here
module os [system] [extern_c] {
       ^
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/module.modulemap:1494:8: error: redefinition of module 'libkern'
module libkern [system] [extern_c] {
       ^
[...]

PR #386, opened 2016-03-26
[CocoaPods] Add SQLite module map for each architecture

cjwirth:
Piggy-backing off of https://github.com/stephencelis/SQLite.swift/pull/380

## ☝️ What does this PR do?

This fixes the build problem where people were unable to build depending on what version of Xcode they were using. I know, it sounds bizarre...

## ✌️ How does it do it? 

I split up the module maps up by architecture, and imported them in the private build settings (basically the same technique I mentioned in [this comment](https://github.com/stephencelis/SQLite.swift/issues/349#issuecomment-199859137))

## 📖  More Information!

While I was investigating, I downloaded both versions of Xcode (dmg and AppStore). The dmg version was fine. The AppStore version would build for devices, but not for the simulator. But! If I rearranged the order of the `x86`, `arm`, `arm64` modules, I could get it to succeed/fail in building in the other way around!

So I thought, I don't know which version of Xcode is correct, but one of them is certainly buggy. But it would be better if it would build for anyone regardless of what version of Xcode they were using.

Yeah, it's a little disappointing to have more files like this to take care of, but I couldn't find another way to make a kind of 'conditional import'. 

Plus! I've gotten it to successfully build on CircleCI, so that make me feel a little bit better! :smile:


acrookston:
This was the first time my build went through on the build server (trying out BuddyBuild). The previous solution only worked locally.

stephencelis:
Round and round we go :)

This is similar to one of the first solutions I had to this problem. The way SwiftPM works makes it a more ideal candidate, too, so :+1:.

stephencelis:
Thanks for investigating! This Xcode update was rockier than I think we all hoped :smile:

cjwirth:
Thanks :smile: 

Yeah I hadn't really followed all the changes as they started happening. I got sick the day after this all started, and just got better in time to pick up the rear end! 

> This Xcode update was rockier than I think we all hoped 

Yeah, no kidding... In more ways than just this library too!

:+1:

fuxx:
Hey :)
Looks good to me, but i am still wondered why this works without the umbrella header.

I try to apply the solution to my own project, but my project requires the umbrella header in the module map.

stephencelis:
@fuxx I believe it's because this module map exists in addition to the generated one, with a _different_ module name (`CSQLite` in this case).

PR #389, opened 2016-03-29
Making small change so that this can play nice with SQLiteCipher.swift

gblotter:
Needed to make `func check()` be public, so that SQLiteCipher.swift can access it in it's extension of the Connection class.

Please let me know what questions you have.  if there is a better way of doing this, I'm open to suggestions.  If you want to check out the details of the PR I have coming for the SQLiteCipher.swift repository, that uses this change, you can check it out here: https://github.com/gblotter/SQLite.swift/tree/SQLiteCipher-play-nice

PR #390, opened 2016-03-31
Created a subspec in your podspec file for SQLiteCipher

gblotter:
The reason why I took this approach is one had to compile the SQLite.swift library to use a codec if they were to get the goodness of SQLCipher, yet the only difference between your other repo and this one was the presence of the Cipher.swift file.

This approach does NOT add the Cipher.swift file to the Xcode project, but rather includes it in the pod's subspec so that if someone chooses to install it, it is present in the build SQLite.swift framework.

If this PR is accepted and merged, than #389 wouldn't be necessary, and neither would the separate git repo for SQLiteCipher.swift (especially https://github.com/stephencelis/SQLiteCipher.swift/pull/12)


pepejeria:
Any update on this?

TheNephilim88:
@stephencelis ?

jberkel:
This needs rebasing since there are now sub specs defined in the current master.

Also not sure if `SWIFT_INCLUDE_PATHS` is needed in the subspec. Otherwise there'll be a mixup of `CSQLite` (defined by the system) and the `SQLiteCipher` framework. 

I'm not sure how SQLiteCipher would integrate, presumably we'd need another compile time flag?

```swift
#if SQLITE_SWIFT_STANDALONE
import sqlite3
#else if  SQLITE_SWIFT_SQLITECIPHER
import SQLiteCipher
#else
import CSQLite
#endif
```

jberkel:
No activity, closing. Please reopen a new PR if needed.

jberkel:
this has been reworked into #553

PR #396, opened 2016-04-04
Restrict use of qualified table names

### LINKED ISSUE #395: Attempts to use qualified table name while creating index
SQLite doesn't accept a qualified table name when creating an index.

```
Table("tbl", database: "main").createIndex([id])
```

will create invalid syntax like

```
CREATE INDEX "main"."index_tbl_on_id" ON "main"."tbl" ("id")
                                         ^^^^^^^
```

jlawton:
Qualified table names are not allowed when creating indexes or foreign key references.

Fixes #395

### DIFF
--- a/SQLite/Typed/Query.swift
+++ b/SQLite/Typed/Query.swift
@@ -744,6 +744,13 @@ extension QueryType {
         ])
     }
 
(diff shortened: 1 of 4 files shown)


stephencelis:
Nice solution! Thank you!

PR #398, opened 2016-04-06
Scalar functions should throw errors

JavaNut13:
I want to be able to use the scalar functions in my app but don't want the risk of crashing due to some unexpected state of the database. The rest of the database queries throw exceptions to be handled by the application, I don't see any reason why this shouldn't apply to scalar functions (if there is and I've missed something really obvious then I'd like to hear it).


MrMage:
+1.

nickmshelley:
@JavaNut13: I think this is a great change. Would you mind merging master manually so the PR can be merged?

nickmshelley:
Sorry for the extra work @JavaNut13, but I had to revert a commit I made recently because it broke concurrent access for single connections (which I think is pretty important). Would you mind merging again?

JavaNut13:
Merged 👍

nickmshelley:
Thanks! I'd love to see this merged @stephencelis if you get a chance.

PR #401, opened 2016-04-07
Embed C Library Directly into project

ryanholden8:
Hi @stephencelis,

I'm not sure the reasons against this (so forgive me please (; ) and it might have already been discussed (couldn't find the discussion) but this would solve a few problems:

- The issues with cocoapods and the sqlite3 header file
- iOS SQLite version is very [much behind](https://github.com/yapstudios/YapDatabase/wiki/SQLite-version-(bundled-with-OS)).
- SQLite 3.8-3.12.0 has [many speed improvements.](http://www.sqlite.org/changes.html)
- FTS5 is becoming more stable and useful (bm25 ranking, speed). [JSON1 extension](https://www.sqlite.org/json1.html) is cool too. If this project really wanted to utilize these features without requiring iOS 11 then embedding the C library would be required anyways.

### DIFF
--- a/CocoaPods/appletvos/module.modulemap
+++ b/CocoaPods/appletvos/module.modulemap
@@ -1,4 +1,0 @@
-module CSQLite [system] {
-    header "/Applications/Xcode.app/Contents/Developer/Platforms/AppleTVOS.platform/Developer/SDKs/AppleTVOS.sdk/usr/include/sqlite3.h"
-    export *
-}
--- a/CocoaPods/appletvsimulator/module.modulemap
+++ b/CocoaPods/appletvsimulator/module.modulemap
@@ -1,4 +1,0 @@
-module CSQLite [system] {
-    header "/Applications/Xcode.app/Contents/Developer/Platforms/AppleTVSimulator.platform/Developer/SDKs/AppleTVSimulator.sdk/usr/include/sqlite3.h"
-    export *
-}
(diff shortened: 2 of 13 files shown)


kdubb:
FMDB does this using a subspec (Cocoapod). Couldn't this be done in this project as well?

kdubb:
FYI, I am completely for this, we've effectively abandoned using the SQLite from iOS because we require a stable version to develop against.

stephencelis:
I definitely like the idea behind this, but I have a few concerns with the current implementation:

- As @kdubb touched upon, can we use a git submodule and CocoaPods subspec instead of embedding everything directly in the repo? Using a git submodule makes it much easier to keep things in sync as the upstream SQLite library sees new releases and keeps the commit diff/noise down.
- I also would prefer not to _require_ every developer to bundle SQLite with their app. I haven't done the math, but I imagine the app payload to increase by a nontrivial amount.

jberkel:
I've tried a few other options but CocoaPods is so far the best way to handle this. #431 adds a subspec which downloads the latest stable version of sqlite and includes it in the build. 

Not sure how this could be done elegantly with submodules / Carthage – it would probably require separate targets (linking/framework) which would complicate the build even further.

jberkel:
SQLite can be integrated via CocoaPods now. closing.

PR #411, opened 2016-04-16
Connection pool

kdubb:
Implement simultaneous multi reader, single write connection pool.  Uses WAL and exclusive VFS mode to achieve concurrency.


SQLite/Core/Connection.swift
```diff
@@ -25,8 +25,209 @@
...
+    case Immediate = "IMMEDIATE"
+    
+    /// Immediately acquires an exclusive lock on all databases.
+    case Exclusive = "EXCLUSIVE"
+    
+}
+
+
+/// Protocol to an SQLite connection
+public protocol Connection {
```
stephencelis: Per the Swift naming guidelines can we make this `ConnectionProtocol` and retain the default `Connection` class name?

SQLite/Core/ConnectionPool.swift
```diff
@@ -0,0 +1,177 @@
...
+
+import Dispatch
+import CSQLite
+
+
+private let vfsName = "unix-excl"
+
+
+/// Connection pool delegate
+public protocol ConnectionPoolDelegate {
```
stephencelis: Do you have examples around how you're using this delegate?

stephencelis:
This is great! Naming/delegate question aside, I'd love to get this merged!

SQLite/Core/Connection.swift
kdubb: Wouldn't `ConnectionType` be more Switfy? (thinking of `SequenceType`, etc)

kdubb: I haven't kept up on the Swift-Evo list as of late. I know they are renaming basically everything.

kdubb: Also, I kind of liked having `DBConnection`, regardless of the protocol name, simply because it has more capabilities than the protocol does and the name kind of explains why in that case.  Although, with a renamed protocol, simply `Connection` works as well.

SQLite/Core/ConnectionPool.swift
kdubb: The protocol informs listeners that a new connection has been added. Remember that each connection is distinct. So  things like `createFunction` must happen against each added connection; not just the pool itself.

This is the method FMDatabasePool used. The thought had crossed my mind to *capture* calls like `createFunction` and perform them automatically as connections are created. That solution is a long term one that will require a bit of thinking because I don't yet know the scope of what could fall into that category.  The delegate is the obvious first step.

SQLite/Core/Connection.swift
stephencelis: Yep, this is the proposal: https://github.com/apple/swift-evolution/blob/master/proposals/0006-apply-api-guidelines-to-the-standard-library.md

> Strip `Type` suffix from protocol names. In a few special cases this means adding a `Protocol` suffix to get out of the way of type names that are primary (though most of these we expect to be obsoleted by Swift 3 language features).

I need to go back and update the other `Type`-suffixed protocols, as well.

kdubb: Given this I'd love to keep it as is.  `DBConnection` & `BorrowedConnection` (although you never use it directly) are more descriptive and seem to be more in line with the eventual Swift naming schemes; the alternative of `ConnectionProtocol`, `Connection` & `BorrowedConnection` seems less desirable to me even if it means a bit of renaming in projects.

In reality, projects shouldn't have to do much renaming at all since `Connection` just becomes a protocol.  Only during initialization & when using special functions would you need to use `DBConnection`.

SQLite/Core/ConnectionPool.swift
stephencelis: Ah, right. That makes sense! I wonder if the same thing could be accomplished internally/privately for most of the connection's helper functions (along with an additional public callback interface). _I.e._ when `createFunction`, etc., are called, that list of setup closures is appended to, called against all existing connections, and the entire list is called against newly-spawned connections.

Something like this:

``` swift
let pool = try ConnectionPool()
pool.foreignKeys = true
// pool.onConnect { db in db.run("PRAGMA foreign_keys = 1") }
```

I can see this being slightly less flexible with the helper function abstraction, but the simple callback block in isolation could be good enough if we don't define those helpers directly on `ConnectionPool`.

My main hesitation with adding a delegate API prematurely is that I just haven't seen the delegate pattern used as much in pure Swift libraries. I wonder if it's the appropriate API and if we should consider alternatives before committing to the public interface.

SQLite/Core/Connection.swift
stephencelis: I wonder if we can find another, happier medium. The reaction I had to the current naming direction is that it's a little harder to derive specific meaning on its own. I'm left wondering why only one of the types is prefixed with the abbreviation `DB` and what the significance of `DB` has over the other types. I also wonder if decorators like `DB` are slightly redundant in a SQLite library?

I actually prefer `ConnectionType` over `ConnectionProtocol`, but was trying to move in the general direction of Swift norms. I wonder if we could come up with a different protocol name that we like even better. Or maybe `DBConnection` could change to something else, like `ConcreteConnection`, though that descriptor still seems slightly confusing to me.

Any other ideas?

kdubb: I agree that 'DB' wasn't the best. Prior to that I had almost settled on `DirectConnection`.  Which may be more to your liking?

stephencelis: Yup! Definitely prefer that! Has very clear meaning on its own.

kdubb:
@stephencelis Updated to use setup closures very close to your suggestion.  Like it much better.  Also renamed to `DirectConnection`

stephencelis:
Looking good! Going to check it out and put it through some tests before merging.

SQLite/Core/ConnectionPool.swift
```diff
@@ -0,0 +1,201 @@
...
+        
+        return writeConnection
+    }
+    
+    // Acquires a read only connection to the database
+    public var readable : Connection {
+        
+        var borrowed : BorrowedConnection!
+        
+        repeat {
```
robertjpayne: Isn't this a busy loop? If all the readers are checked out it would appear this method will just keep repeating busily until one is available. Would be a lot better to use a `dispatch_semaphore` or `NSCondition` to let it sleep and get a signal when a connection is actually available.

SQLite/Core/ConnectionPool.swift
```diff
@@ -0,0 +1,201 @@
...
+            
+            dispatch_sync(lockQueue) {
+                
+                // Ensure database is open
+                self.writable
```
robertjpayne: self.writable isn't thread safe because `var writeConnectionInit` isn't thread safe. While very unlikely this could cause a crash.

SQLite/Core/ConnectionPool.swift
```diff
@@ -0,0 +1,201 @@
...
+                
+                let connection : DirectConnection
+                
+                if let availableConnection = self.availableReadConnections.popLast() {
+                    connection = availableConnection
+                }
+                else {
+        
+                    let flags = SQLITE_OPEN_READONLY | SQLITE_OPEN_WAL | SQLITE_OPEN_NOMUTEX
```
robertjpayne: I'd implement a cap on read connections, the general idea of a database pool is that it has a reader cap and if they are all busy the next caller will wait.

kdubb: Yep, this was accidentally removed when the delegate was removed.  Need to implement connection limits.

SQLite/Core/ConnectionPool.swift
kdubb: To be honest I need to look at this again... This seems to be a holdover from an earlier version/feature.  This code should find an available connection, wait for one or fail.  None of that requires a loop. Good catch!

robertjpayne: Right now because it will always create a new reader it'll never busy loop, but if it didn't it would create a busy loop just need a signal of some sort…

You also want to make sure that signal isn't does block the `lockQueue` otherwise a connection wont be able to be checked in.

nickmshelley:
@kdubb: Is there something I can do to help get this through? I don't want to step on any toes but I'd love to see this in and am willing to finish up whatever's remaining if you're too busy at the moment.

SQLite/Core/Connection.swift
```diff
@@ -67,10 +268,32 @@ public final class Connection {
     ///     Default: `false`.
     ///
     /// - Returns: A new database connection.
-    public init(_ location: Location = .InMemory, readonly: Bool = false) throws {
+    public convenience init(_ location: Location = .InMemory, readonly: Bool = false, vfsName: String? = nil) throws {
```
nickmshelley: Looks like the function docs above need to be updated to include the `vfsName` parameter.

nickmshelley:
I opened https://github.com/stephencelis/SQLite.swift/pull/451 because this doesn't seem to be going anywhere. Let me know if there's a better way to get this through than opening a new PR.

SQLite/Core/ConnectionPool.swift
nickmshelley: @robertjpayne Can you elaborate a bit? I couldn't find anything saying `dispatch_once_t()` isn't thread safe, so I'm not sure what to fix.

SQLite/Core/ConnectionPool.swift
nickmshelley: I did this in https://github.com/stephencelis/SQLite.swift/pull/451/commits/960309eb02724ae0f18f721942fb22473bb4b602 and would appreciate any feedback.

SQLite/Core/ConnectionPool.swift
nickmshelley: I did this in https://github.com/stephencelis/SQLite.swift/pull/451/commits/960309eb02724ae0f18f721942fb22473bb4b602 and would appreciate any feedback.

kdubb:
@nickmshelley My apologies I hadn't seen the resurrection of this PR until I was just notified about it being merged.  All worked out it seems!

nickmshelley:
@kdubb I hope so. Sorry if I stole any thunder.

nickmshelley:
Also thanks for doing the bulk of the work @kdubb.

stephencelis:
Thanks for shepherding this, @nickmshelley!

PR #412, opened 2016-04-17
Set project tab & indent settings

kdubb:
Makes life easier so when we switch into this project Xcode automatically selects the correct tab & indent width.


stephencelis:
Thanks!

PR #416, opened 2016-04-21
Performance improvements for Blob

robertjpayne:
We use Blobs a lot in our database and find the current API to not allow much performance tuning. This pull request address this by:

1. Blobs are now a class reference type rather than a value type so it's internal bytes are never copied.
2. Blobs initialiser now uses a more similar NSData like api; `init(bytes: UnsafePointer<Void>, length: Int, freeWhenDone: Bool = false)`. This allows zero-copy blobs by allowing them to assume ownership of existing memory chunk without copying it.
3. NSData when running under Apple platforms now utilise the objc runtime's associated objects to attach themselves to a blob and ensure the blob uses the byte buffer but does not copy the bytes.

There is minimal impact on existing code, most users were probably using the NSData <-> Blob conversion anyways and if it is a big concern it wouldn't be too hard to re-add the old bytes initialiser.

The only pitfall I can see is if a user grabs a Blob from a NSMutableData object, I could potentially put some checks in place for this?


robertjpayne:
Actually there's a few bugs in this implementation, why might make more sense is to on Apple Platforms typealias a Blob as NSData, once Swift 3.0 drops with full Foundation support it'll work for Linux as well?

robertjpayne:
Not sure what's up with Travis, it's pulling an old commit that I squashed.

PR #426, opened 2016-05-09
Fix transactions not being rolled back when the individual statements succeed, but the committing the transaction fails.

MrMage:
This can happen e.g. when using deferred foreign keys (as seen in the tests), or when accessing the database from multiple processes, causing a commit to fail with an "database is busy" error.


MrMage:
I have merged the latest changes and all tests pass for me. Could you please have a look at this pull request?

jberkel:
merged into release branch

PR #431, opened 2016-05-12
CocoaPods: build standalone version of SQLite with subspec

jberkel:
Adresses #401.

Not sure about the "standalone" spec name – I just followed the pattern used in https://github.com/ccgus/fmdb/blob/master/FMDB.podspec.

Passes `pod spec lint --allow-warnings`.

The existing behavior and setup is left unchanged. I can't think of a good way to integrate the SQLite sources in a submodule, maybe this could be addressed in a separate PR.

It also needs tests – maybe a dummy project could be set up automatically, including test suite.

### DIFF
--- a/Documentation/Index.md
+++ b/Documentation/Index.md
@@ -111,8 +111,26 @@ install SQLite.swift with Carthage:
 
  3. Run `pod install`.
 
+ #### Requiring a specific version of SQLite
+
+ If you want to use a more recent version of SQLite than what is provided with the OS you can require the `standalone` subspec:
+
+``` ruby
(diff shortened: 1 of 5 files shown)


stephencelis:
Nice! This looks good, but I wonder if we can somehow retain the `CSQLite` module name? We were trying to adhere to the SwiftPM System Module guidelines: https://github.com/apple/swift-package-manager/blob/master/Documentation/SystemModules.md#system-modules

jberkel:
Ah, I wondered about the weird naming, now it makes sense. I changed it so that the renaming is avoided, it also makes it clearer which module is being used.

stephencelis:
Beautiful, thanks!

PR #433, opened 2016-05-13
[CocoaPods] integration tests

jberkel:
Tests changes introduced in #431.

Runs tests against standard and standalone subspecs, for iOS/OSX/tvOS. The FTS4 integration test needs to be disabled for standalone builds since newer versions of SQLite need to explicitly enable custom tokenizers: 

https://www.sqlite.org/compile.html#enable_fts3_tokenizer


stephencelis:
😄 Excellent! Thanks for rounding things out with coverage!

Issue #164, opened 2015-07-29
External Content FTS4 Table


go1t:
Is there a way to do this in SQLite.swift? : 
```javascript
 CREATE VIRTUAL TABLE table USING fts4(content='some_other_table', id, name)
```
I'd like to pre-populate my fts4 virtual table, but inserting one by one is very slow since my table contains over 30,000 rows. Is there any recommended way for this in SQLite.swift?

stephencelis:
Many of the options for `fts4` aren't exposed yet but should be. Making this function public will get you closer:

https://github.com/stephencelis/SQLite.swift/blob/89d16192bb73bf179c08c30227fbd9e52be6bf32/SQLite/FTS.swift#L36

Or you can execute the raw SQL instead.

igodzp:
That's one quick reply, thanks! 

I have a bit of the trouble about executing raw SQL. I currently have one read-only database and the read-write database for creating virtual table. The read-only one contains the records I want to pre-populate the virtual table in another database. The location of that read-only database is in the Supporting Files folder inside the project. How could I possibly use the table from that database in raw SQL? I can open it with SQLite.swift database with these command just fine: 

```swift
var path = NSBundle.mainBundle().pathForResource("db", ofType: "sqlite3")!
let readonlydb = Database(path, readonly: true)
```
but I cant seem to be able to access the database and tables with raw SQL.  Thank you in advance!

lukescott:
It would also be nice if there was a way to setup the triggers to keep the data in sync, as shown here: http://www.sqlite.org/fts3.html#section_6_2_2 .

mikemee:
Added to [Feature Requests](https://github.com/stephencelis/SQLite.swift/blob/master/Documentation/Planning.md).

PR #435, opened 2016-05-22
Expose full FTS4/5 configuration

jberkel:
This PR adds support for setting all available parameters for FTS4 with a fluent builder-style interface, #164.

Bonus: Support for [FTS5](https://www.sqlite.org/fts5.html)  which differs in a lot of ways from FTS4 but also shares a few config options.

However most users won't need this bit so it might make sense to compile it in conditionally? OTOH it's not a lot of extra code, and pulling in a FTS5-enabled version of SQLite is fairly straightforward now (with CocoaPods / #431).

### DIFF
--- a/Documentation/Index.md
+++ b/Documentation/Index.md
@@ -1399,6 +1399,22 @@ try db.run(emails.create(.FTS4([subject, body], tokenize: .Porter)))
 // CREATE VIRTUAL TABLE "emails" USING fts4("subject", "body", tokenize=porter)
 ```
 
+We can set the full range of parameters by creating a `FTS4Config` object.
+
+``` swift
+let emails = VirtualTable("emails")
+let subject = Expression<String>("subject")
+let body = Expression<String>("body")
+let config = FTS4Config()
+    .column(subject)
+    .column(body, indexed: false)
+    .languageId("lid")
+    .order(.Desc)
+
+try db.run(emails.create(.FTS4(config))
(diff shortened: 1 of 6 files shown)


stephencelis:
This is really cool! And makes me want to revisit the `CREATE TABLE` builder for a chained pattern like this.

One consideration I've had for `column()` options is to use an options set instead of named parameters (since a lot of them are just `true` or not). Any thoughts on moving that direction? _E.g._, with your builder pattern applied to the current table builder:

``` swift
try db.run(users.create()
    .column(id, [.primaryKey])
    .column(name)
    .column(email, [.unique]))
```

jberkel:
Good point, that should also simplify the method signatures quite significantly. `TableBuilder` has currently almost 40 declared public methods.

Documentation/Index.md
```diff
@@ -1414,6 +1430,22 @@ let replies = emails.filter(subject.match("Re:*"))
...
+When linking against a version of SQLite with [FTS5](http://www.sqlite.org/fts5.html) enabled we can create the virtual table
+in a similar fashion.
+
+```swift
+let emails = VirtualTable("emails")
+let subject = Expression<String>("subject")
+let body = Expression<String>("body")
+let config = FTS5Config()
+    .column(subject)
+    .column(body, indexed: false)
```
stephencelis: In the spirit of my previous comment, what do you think of updating this to be:

``` swift
    .column(body, [.unindexed])
```

stephencelis:
This is great, thanks again!

PR #436, opened 2016-05-23
I have a database where there can be an empty ( non nil ) date string

aleinsight:
A better solution would be to either provide a default value for the case when the string is not a well formed date or to make fromDatatypeValue() return an optional NSDate and return nil if anything goes wrong.

### DIFF
--- a/SQLite/Foundation.swift
+++ b/SQLite/Foundation.swift
@@ -47,7 +47,7 @@ extension NSDate : Value {
     }
 
     public class func fromDatatypeValue(stringValue: String) -> NSDate {
-        return dateFormatter.dateFromString(stringValue)!
+        return dateFormatter.dateFromString(stringValue) ?? NSDate()
     }
 
     public var datatypeValue: String {


stephencelis:
Hm, `fromDataValue()` should only be called if the data is not `nil`. Is your column of type `Expression<NSDate?>`?

aleinsight:
Hi Stephen,

I can confirm that the expression is Expression<NSDate?> (with optional NSDate) and fromDatatypeValue(stringValue: String) gets called with an empty string ( not nil string ).

In the database I have an empty string (“"), not a nil string.

dateFormatter.dateFromString returns an optional date, in case the date string is not well formed such as in my case, so using ! to unwrap the result of the call ends up trying to unwrap a nil value.

That is the reason why the code does crash.

However I cannot change that specific database and I think that others may have the same issue at some point.

Thanks,
Alessandro

On 25 May 2016, at 14:02, Stephen Celis <notifications@github.com<mailto:notifications@github.com>> wrote:


Hm, fromDataValue() should only be called if the data is not nil. Is your column of type Expression<NSDate?>?

SQLite/Foundation.swift
```diff
@@ -47,7 +47,7 @@ extension NSDate : Value {
     }
 
     public class func fromDatatypeValue(stringValue: String) -> NSDate {
-        return dateFormatter.dateFromString(stringValue)!
+        return dateFormatter.dateFromString(stringValue) ?? NSDate()
```
nickmshelley: The problem with this is that it makes it seem like the database is actually storing a string that represents the current date, which can be misleading and confusing to users.

aleinsight: Hi, a cleaner way to do this could be:

    public class func fromDatatypeValue(stringValue: String) -> NSDate? {
        return dateFormatter.dateFromString(stringValue)
    }

On 19 Jul 2016, at 21:21, Nick Shelley <notifications@github.com<mailto:notifications@github.com>> wrote:


In SQLite/Foundation.swift<https://github.com/stephencelis/SQLite.swift/pull/436#discussion_r71415629>:

> @@ -47,7 +47,7 @@ extension NSDate : Value {
>      }
>
>      public class func fromDatatypeValue(stringValue: String) -> NSDate {
> -        return dateFormatter.dateFromString(stringValue)!
> +        return dateFormatter.dateFromString(stringValue) ?? NSDate()


The problem with this is that it makes it seem like the database is actually storing a string that represents the current date, which can be misleading and confusing to users.

jberkel:
out-of-date, closing

PR #437, opened 2016-05-25
Add missing imports

p2:
Adds two imports that become apparent when not compiling the original project, but adding the files directly.

### DIFF
--- a/SQLite/Core/Connection.swift
+++ b/SQLite/Core/Connection.swift
@@ -22,6 +22,7 @@
 // THE SOFTWARE.
 //
 
+import Foundation
 import Dispatch
 #if SQLITE_SWIFT_STANDALONE
 import sqlite3
--- a/SQLite/Core/Connection.swift
+++ b/SQLite/Core/Connection.swift
@@ -22,7 +22,7 @@
 // THE SOFTWARE.
 //
 
-import Foundation
+import Foundation.NSUUID
 import Dispatch
 #if SQLITE_SWIFT_STANDALONE
 import sqlite3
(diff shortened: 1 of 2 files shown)


SQLite/Core/Connection.swift
```diff
@@ -22,6 +22,7 @@
 // THE SOFTWARE.
 //
 
+import Foundation
```
jberkel: the missing dependency here is NSUUID
(I wonder if it's worth making this explicit, instead of doing these catch-all imports)?

```swift
import Foundation.NSUUID
```

SQLite/Typed/CoreFunctions.swift
```diff
@@ -22,6 +22,9 @@
 // THE SOFTWARE.
 //
 
+import Foundation
```
jberkel: likewise, `NSData`

SQLite/Core/Connection.swift
p2: That would be an option. In case of Foundation I wouldn't think that's necessary, it's highly likely all of Foundation are imported elsewhere anyway.

jberkel: Not as an optimization, more as a documentation of the dependencies of the file. But it's not commonly done in Swift/ObjC.

stephencelis:
I appreciate @jberkel's suggestions. Any chance you can make the changes before we merge?

iNoles:
With these changes, there is no needed for @import Foundation in SQLite.h then?
https://github.com/stephencelis/SQLite.swift/blob/master/SQLite/SQLite.h

jberkel:
don't think so, you need it for the `FOUNDATION_EXPORT` macro

p2:
Sure, works for me, updated.

stephencelis:
Thanks!

PR #440, opened 2016-05-28
Use SQLITE_OPEN_NOMUTEX for readonly database to avoid locking

stephanheilner:
According to the SQLite docs, if the SQLITE_OPEN_NOMUTEX flag is set, then the database connection opens in the multi-thread threading mode. This change will make it so that readonly database connections will have muti-threaded access since there's no risk of trying to write while reading. It keeps SQLITE_OPEN_FULLMUTEX for write databases.

Docs here: https://www.sqlite.org/c3ref/open.html

### DIFF
--- a/Sources/SQLite/Core/Connection.swift
+++ b/Sources/SQLite/Core/Connection.swift
@@ -151,27 +151,53 @@ public final class Connection {
     ///
     ///     Default: `.inMemory`.
     ///
+    ///   - flags: sqlite3_open_v2 option flags.
+    ///
+    /// - Returns: A new database connection.
+    public init(_ location: Location = .inMemory, flags: [Flag]) throws {
+        var flags = flags
+        let firstFlag = flags.remove(at: 0)
+        let sqlite3Flags = flags.reduce(firstFlag.rawValue) { $0 | $1.rawValue }
+        
+        try check(sqlite3_open_v2(location.description, &_handle, sqlite3Flags, nil))
+        queue.setSpecific(key: Connection.queueKey, value: queueContext)
+    }
+    
+    /// Initializes a new SQLite connection.
+    ///
+    /// - Parameters:
+    ///
+    ///   - location: The location of the database. Creates a new database if it
+    ///     doesn’t already exist (unless in read-only mode).
+    ///
+    ///     Default: `.inMemory`.
+    ///
     ///   - readonly: Whether or not to open the database in a read-only state.
     ///
     ///     Default: `false`.
     ///
     /// - Returns: A new database connection.
-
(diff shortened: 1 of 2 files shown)


groue:
The "multi-thread" mode has a misleading name, because it is not thread-safe at all. Quoting https://www.sqlite.org/threadsafe.html:

> The SQLITE_OPEN_NOMUTEX flag causes the database connection to be in the multi-thread mode.
>
> Multi-thread: In this mode, SQLite can be safely used by multiple threads provided that no single database connection is used simultaneously in two or more threads.

This does not exactly give readonly databases multi-threaded access.

On a side note, the topic of concurrent accesses to a SQLite databases has been solved, tested, and stress-tested in [GRDB](https://github.com/groue/GRDB.swift#database-connections).

jberkel:
@groue GRDB looks interesting, it should probably be added to https://github.com/stephencelis/SQLite.swift#alternatives (#441). stress-tested = used in production apps?

groue:
> GRDB looks interesting

Thanks for having a look!

> stress-tested = used in production apps?

Good question :-) I can't talk for others. I've only shipped GRDB in private OS X apps - the iOS apps are still in the making. My [own](https://github.com/groue/GRDB.swift/blob/master/Tests/Public/Core/DatabasePool/DatabasePoolConcurrencyTests.swift) [tests](https://github.com/groue/GRDB.swift/blob/master/Tests/Public/Core/DatabaseQueue/DatabaseQueueConcurrencyTests.swift) put sqlite and GRDB in delicate threading scenarios. The [GRDBDemo](https://github.com/groue/GRDBDemo) sample code stress-tests GRDB and its [FetchedRecordsController](https://github.com/groue/GRDB.swift#fetchedrecordscontroller) by spawning dozens of concurrent writes and read while maintaining a UITableView synchronized with the database, in an animated fashion. I also know of one user who [could not](https://github.com/groue/GRDB.swift/issues/60#issuecomment-221438841) make GRDB crash. Finally, GRDB takes concurrency quite seriously, by design: https://medium.com/@gwendal.roue/four-different-ways-to-handle-sqlite-concurrency-db3bcc74d00e

stephanheilner:
Thread safety isn't a huge risk because this only applies to readonly database connections. Thread safe for a database would just mean its free of race conditions when accessed by multiple threads simultaneously. If you're not writing to the database, what race condition could there be? Since this pull request is for SQLite.swift and not for GRDB, there's still value in merging this change.

jberkel:
See comment in #354

stephanheilner:
I made the changes you recommended in https://github.com/stephencelis/SQLite.swift/pull/354

PR #441, opened 2016-05-30
Add GRDB to list of alternatives

jberkel:


stephencelis:
Ah, I meant to do this awhile back. We should also probably prune inactive projects (ones that weren't updated for Swift 2).

Issue #438, opened 2016-05-25
Crash when lastInsertRowId is negative


zmeyc:
Consider, for example, the following statement:

`Insert(template: "INSERT INTO \"sessions\" (\"owner_id\", \"started\", \"victim_name\", \"topic\", \"line\", \"chat_id\") VALUES (?, ?, ?, ?, ?, ?)", bindings: [Optional(230594948), Optional(1), Optional(""), Optional(""), Optional(0), Optional(-147636081)])`

chat_id is INTEGER PRIMARY KEY and negative value is inserted.

sqlite3_last_insert_rowid() returns -147636081 (as expected)

Lib's lastInsertRowId turns it into nil:

```
    public var lastInsertRowid: Int64? {
        let rowid = sqlite3_last_insert_rowid(handle)
        return rowid > 0 ? rowid : nil
    }

```

The app crashes because Connection.run() dereferences nil.

zmeyc:
I'm not sure if any return value of sqlite3_last_inserted_rowid can be treated as an error condition. Lib's wrapper probably shouldn't return an optional value:
https://www.sqlite.org/c3ref/last_insert_rowid.html

iNoles:
Well, it can fail by any constraint violation.

iNoles:
1 )  "If no successful INSERTs into rowid tables have ever occurred on the database connection D, then sqlite3_last_insert_rowid(D) returns zero."
2) constraint violation = also return zero.

zmeyc:
>An INSERT that fails due to a constraint violation is not a successful INSERT and does not change the value returned by this routine.

I think it's zero initially, but changed only on successful inserts.

stephencelis:
Would you like to submit a PR that changes that line to the following?

``` swift
return rowid != 0 ? rowid : nil
```

zmeyc:
@stephencelis 
It appears zero rowid is a valid value, I've done some testing:

```
sqlite> CREATE TABLE foo (id INTEGER PRIMARY KEY, bar TEXT); 
sqlite> SELECT last_insert_rowid();
0
sqlite> INSERT INTO foo (id, bar) VALUES (0, 'test');
sqlite> SELECT last_insert_rowid();
0
sqlite> INSERT INTO foo (bar) VALUES ('more'); 
sqlite> SELECT last_insert_rowid();
1
sqlite> SELECT * FROM foo; 
0|test
1|more
sqlite> SELECT rowid, * FROM foo; 
0|0|test
[...]

PR #442, opened 2016-06-01
Prevent from crash when lastInsertRowId is negative

FrainL:
Really need to use negative primary key in my project.


stephencelis:
Thanks! This is a good temporary fix for https://github.com/stephencelis/SQLite.swift/issues/438

PR #458, opened 2016-06-23
Closes #457: Split up tests into more Travis jobs.

### LINKED ISSUE #457: Split up tests better in travis
Currently Travis runs two jobs, an iOS one and a Mac one. The majority of time spent on both of these jobs is in the cocoapods tests, which test all platforms (iOS, tvOS, watchOS, and macOS) in both jobs.

I think ideally there would be a job or two for unit tests, and then probably a cocoapods test job for each platform. This would keep us from running into the 50 minute Travis time limit for Mac jobs.

nickmshelley:


nickmshelley:
In verifying that I did it right, I noticed that Mac tests are running regardless of the scheme set in the environment, and it seems like that's been the behavior for a while. I think it's due to this line: https://github.com/stephencelis/SQLite.swift/blob/master/Makefile#L2. I figure I might as well fix that as part of this PR.

stephencelis:
This is great, thanks! Much-needed infrastructure improvements.

PR #459, opened 2016-06-23
Run desired unit tests instead of only Mac.

nickmshelley:
If you look at the logs of any of the previous travis builds, they all run Mac tests, even when the scheme is set to "SQLite iOS". This should fix that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 BUILD_TOOL = xcodebuild
-BUILD_SCHEME = SQLite Mac
-BUILD_ARGUMENTS = -scheme "$(BUILD_SCHEME)"
+ifeq ($(BUILD_SCHEME),SQLite iOS)
+	BUILD_ARGUMENTS = -scheme "$(BUILD_SCHEME)" -sdk iphonesimulator
+else
+	BUILD_ARGUMENTS = -scheme "$(BUILD_SCHEME)"
+endif
 
 XCPRETTY := $(shell command -v xcpretty)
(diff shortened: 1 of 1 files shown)


nickmshelley:
I can't see anything helpful in the logs about why that one job failed. I'm hoping it was just a spurious failure, so I'll close and reopen to retrigger the tests.

nickmshelley:
Yep, looks like it passed. Weird.

Makefile
```diff
@@ -1,6 +1,9 @@
 BUILD_TOOL = xcodebuild
-BUILD_SCHEME = SQLite Mac
```
jberkel: I think it would be useful to leave this default in so that (fast) tests can be run from the command line with a simple `make` or `make test`, instead of a lengthy `make test BUILD_SCHEME='SQLite Mac'`.

nickmshelley: It wasn't a default, it was just ignoring whatever you passed in and always using `SQLite Mac` for the scheme.

But yes, I agree that having a default to make it easier to run quick tests from the command line would be a good change to make.

nickmshelley: I think it would need to be wrapped in an `if [ -z "$BUILD_SCHEME" ]` to make it act like a default instead of an override.

jberkel: it should work via make's conditional =? operator (or ?=, can't remember the order)


Il giorno 8 luglio 2016 @ 16:16:27, Nick Shelley (notifications@github.com(mailto:notifications@github.com)) ha scritto:

Issue #414, opened 2016-04-17
Pluck does not throw


PranKe01:
The pluck function does not throw an exception, but in the documentation you use the try keyword:
let user = try db.pluck(query)

Also in the documentation the menu point "Migrations and Schema Versioning" is not linked.

t4ec:
+1 

Currently I try to check 1 field in DB to make sure it is valid, in case it is not valid app crashes and "try" cannot be used as pluck does not through.

PR #470, opened 2016-07-16
Throw an error occurred inside Query.pluck()

tasanobu:
I fixed the issue reported in #414 to avoid app crash.

### DIFF
--- a/SQLite/Typed/Query.swift
+++ b/SQLite/Typed/Query.swift
@@ -954,11 +954,8 @@ extension Connection {
         return V.fromDatatypeValue(value)
     }
 
-    public func pluck(query: QueryType) -> Row? {
-        guard let rows = try? prepare(query.limit(1, query.clauses.limit?.offset)) else {
-            return nil
-        }
-        return rows.generate().next()
+    public func pluck(query: QueryType) throws -> Row? {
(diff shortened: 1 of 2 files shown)


SQLite/Typed/Query.swift
```diff
@@ -955,7 +955,10 @@ extension Connection {
     }
 
     public func pluck(query: QueryType) -> Row? {
-        return try! prepare(query.limit(1, query.clauses.limit?.offset)).generate().next()
+        guard let rows = try? prepare(query.limit(1, query.clauses.limit?.offset)) else {
+            return nil
```
nickmshelley: We should probably propagate the error up instead of returning `nil`.

nickmshelley: So basically I think this should be

``` swift
public func pluck(query: QueryType) throws -> Row? {
  return try prepare(query.limit(1, query.clauses.limit?.offset)).generate().next()
}

tasanobu:
@nickmshelley
I’ve changed the implementation to propagate an error instead of returning nil

nickmshelley:
Thanks @tasanobu!

@stephencelis I'd love to see this merged when you get some time.

Issue #337, opened 2016-01-21
[Question] Can you please update the documentation?


jlevesque:
The title says it all update your doc.

I've wasted 2 hours on a creation of a custom type. I've searched in the issues for my problem, and I'm not the only with it.

```
extension NSData : Value {

    typealias Datatype = NSData

    static var declaredDatatype: String {
        return Blob.declaredDatatype
    }

    static func fromDatatypeValue(blobValue: Blob) -> Self {
        return self.init(bytes: blobValue.bytes, length: blobValue.bytes.count)
    }

    var datatypeValue: Blob {
        return Blob(bytes: bytes, length: length)
    }
}
```

Type 'NSData' does not conform to protocol Value

And notice that im using 

```
 blobValue.bytes.count
```

Instead of 

```
 blobValue.length
```

Because the doc is not updated.

Please update your the doc, I'm sure that I'm not the only one who has waste his time.

mikemee:
Hi Julien, that must be a change from an older version. Thanks for pointing it out and sorry for the pain!

This is an all volunteer unpaid effort (thanks Stephen). There is no company sponsor, so it's just people like you and me, contributing for all. 

So, it would be fantastic if you could go and edit the documentation with this info and do a Pull Request. GitHub is making that increasingly easy and it would save everyone some time - and especially the future readers. 

Regardless, thanks for the report!

cuddergambino:
Hey,

Ran into the same issue. Thanks for finding the fix @mikemee and saving everyone else the time!

cheers

carlosen14:
Thanks for the info to solve one problem, i spent a couple of days trying to solve by myself.

What about this?

```
extension UIImage {
  public class var declaredDatatype: String {
    return Blob.declaredDatatype
  }
  public class func fromDatatypeValue(blobValue: Blob) -> UIImage {
    return UIImage(data: NSData.fromDatatypeValue(blobValue))! // ->ERROR "Ambiguous use of 'fromDatatypeValue' "
  }
  public var datatypeValue: Blob {
    return UIImagePNGRepresentation(self)!.datatypeValue
  }
}
```

i got this error, and solved deleting my NSData extension, because there is one already defined in SQLite.swift/Foundation.swift

thanks!

jberkel:
no longer applies, closing

PR #471, opened 2016-07-19
Fixed NSData to Blob bridge extension

cuddergambino:
Related to issue #337

Sqlite.Blob.length is deprecated, need to use Sqlite.Blob.bytes.count instead

### DIFF
--- a/Documentation/Index.md
+++ b/Documentation/Index.md
@@ -1178,12 +1178,12 @@ Any object that can be encoded and decoded can be stored as a blob of data in SQ
 We can create an `NSData` bridge rather trivially.
 
 ``` swift
-extension NSData: Value {
+extension NSData{
     class var declaredDatatype: String {
         return Blob.declaredDatatype
     }
     class func fromDatatypeValue(blobValue: Blob) -> Self {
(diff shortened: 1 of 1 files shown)


Documentation/Index.md
```diff
@@ -1178,12 +1178,12 @@ Any object that can be encoded and decoded can be stored as a blob of data in SQ
 We can create an `NSData` bridge rather trivially.
 
 ``` swift
-extension NSData: Value {
+extension NSData{
```
nickmshelley: This should still conform to the `Value` protocol

cuddergambino: Confirming to the value protocol throws an error in Swift 2.2 and up. Says
you can't redefine the inheritance

nickmshelley: Yeah it looks like it's because it's already implemented: https://github.com/stephencelis/SQLite.swift/blob/master/SQLite/Foundation.swift#L27

I think the right fix is to actually update the docs to point to existing implementations, and leave the UIImage example for when you need to add your own custom ones.

jberkel:
no longer applies, closing

PR #451, opened 2016-06-16
Implement connection pool.

nickmshelley:
Most of this work was done in https://github.com/stephencelis/SQLite.swift/pull/411, which seems to have been abandoned, so I'm trying to get it through. Let me know if there's a better way than opening a new PR.

PR #472, opened 2016-07-19
Revert "Implement connection pool."

nickmshelley:
Reverts stephencelis/SQLite.swift#451

Looks like this introduced some single-connection concurrency issues that should probably be worked through first.


nickmshelley:
I'm going to merge this because it's just reverting something I broke. Then I'll leave the merging to the professionals...

PR #473, opened 2016-07-19
Add test for single connection across multiple threads.

nickmshelley:


nickmshelley:
This should fail tests until https://github.com/stephencelis/SQLite.swift/pull/472 is through, which is what broke this functionality.

nickmshelley:
Hm... this fails locally every time...

nickmshelley:
Travis is having issues at the moment. I'll try to see if I can get this to fail on Travis when it's happy again.

I'm worried that the tests run on a single-core machine/VM, so it might be hard to get it to fail consistently on Travis, whereas locally it fails immediately for me.

jberkel:
Merged into release

PR #477, opened 2016-07-28
Update Index.md

Fershark:
Added sample code of how to successfully copy and use bundle "seed" database for writing in an app


jberkel:
thanks for your contribution, but out of scope

Issue #445, opened 2016-06-07
tvOS target not supported in app extensions


Tableau-David-Potter:
I'm working on integrating this framework into one of my frameworks which we need to work in app extensions.  When building for tvOS, I get a warning saying that I'm "linking against a dylib which is not safe for use in application extensions."  If I check the checkbox on the General tab for the tvOS target, the problem goes away.  Is this safe to do?  If so, could this be changed in the repo?

Thanks,
**_David_**

stephencelis:
This should probably be added to the repo proper!

Tableau-David-Potter:
I submitted a pull request with a fix for this:  https://github.com/stephencelis/SQLite.swift/pull/478

Tableau-David-Potter:
When do you think you can accept this pull request?

jberkel:
change has been merged. closing

Tableau-David-Potter:
Thanks.

PR #478, opened 2016-07-28
Allow the framework to be included in app extensions for tvOS

Tableau-David-Potter:


Tableau-David-Potter:
Is there a plan to accept this merge request?

jberkel:
Not sure how app extensions are supposed to work, but isn't this a setting which is very use case specific? How do you integrate SQLite.swift in your project? Maybe you could override it in your project settings?

Tableau-David-Potter:
@jberkel This setting is required on all frameworks that are included in an extension app, such as a Today widget extension or a watch extension.

jberkel:
Well that was point – the users who actually need to set it can probably do so in their own project?

Tableau-David-Potter:
@jberkel Please try including this framework into a tvOS project and tell me what you see.  When I do this I get a warning telling me that I'm "linking against a dylib which is not safe for use in application extensions."  You can see my report in issue #445.

jberkel:
How do you integrate the framework in your project?

Tableau-David-Potter:
Using Carthage.

jberkel:
Ah ok, in which case you can't easily change the config. Also this setting already exists in the other targets so we should apply it consistently. Can you update the branch so it can get merged?

PR #481, opened 2016-07-29
Argument count fixed for binary custom functions.

andrei-v-frolov:


jberkel:
Can you add some test cases for your changes?

andrei-v-frolov:
Will do.

PR #488, opened 2016-08-11
Update Index.md

Fershark:
Added code to show how to successfully copy and use a bundled "seed" database for writing in an app.


jberkel:
I don't think the documentation is the right place for that sort of code – it's rather generic and not really related to the project itself.

Fershark:
I though it could be useful, because in the documentation it said that:
"We welcome sample code to show how to successfully copy and use a bundled "seed" database for writing in an app."

stephencelis:
Thanks for the submission! I tend to agree with @jberkel, though. This does seem more general in scope than SQLite.swift-specific documentation. Perhaps it would be a good fit for Stack Overflow Documentation?

PR #493, opened 2016-09-14
compatibility with carthage & xcode 8

leoschweizer:
Hi,

this is the minimal changeset I had to apply to get SQLite.swift working again after upgrading to XCode 8 when included as Carthage dependency.

You might want to include this because right now, if you are on XCode 8, you can't use SQLite.swift (at least through Carthage).

I understand that there are probably more changes required to make it work with CocoaPods and the like, so if you don't want to make this partial change feel free to close this PR 😉


jberkel:
this is already in progress: #475

leoschweizer:
@jberkel I'm aware of that, but unfortunately the `xcode8` branch does not build with Carthage (at least for the tvos target). Unfortunately my change is probably not compatible to XCode 7 as the Travis result suggests, so I'll just stick with my fork for know until things are sorted out.

jberkel:
ok, what exactly is the change needed for carthage?

leoschweizer:
@jberkel I've created another PR on the xcode 8 branch, just a minor change but it seems to work fine now, see #494

PR #499, opened 2016-09-15
fixed system sqlite3 lib path for Xcode 8

mariotaku:
Current version linked to MacOSX10.11.sdk, when build in Xcode 8 it will throw errors.

This will make SQLite.swift build-able with new toolchains.


jberkel:
Thanks for the PR! However this should already fixed in the xcode8 branch.

PR #501, opened 2016-09-16
Fully Migrated to Swift 3

mariotaku:
This version works fine in my project and passed all unit tests.

You may want to create a new branch based on `xcode8` branch.

Issue #492, opened 2016-09-12
Carthage SQLite.framework not portable to different systems: CSQLite missing


maun:
When I build the framework using carthage everything works as expected locally, but on other systems I get this error:
```
error: missing required module 'CSQLite'
```
Rebuilding with carthage on these systems fixes the error, but the framework then does not work on the first system anymore.

jerrywell:
+1 for Xcode 8.0 GM

maun:
This was with 7.3.1

jberkel:
ah, I actually misread your bug report. weird, is the module loaded on runtime? can you explain what exactly you mean by "on other system"? do you take the compiled framework and include it another app, on a different machine?

maun:
We have a project checked in to git including the compiled framework. It is imported using the normal import statement in swift. By my system I meant my computer with my os, my coworker uses his mac.

lyc2345:
Same here, I think I misopen my project with Xcode 8, then I open Xcode 7.3, then run the project, error appeared.
header "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk/usr/include/sqlite3.h" not found

And In Connection.swift
Could not build Objective-C module 'CSQLite'

tuchangwei:
@lyc2345  There is a tmp fix - You can change your Xcode app name `Xcode7.3` to `Xcode` to solve this issue.

maun:
Both our Xcodes where named Xcode.app, still got the error.

jberkel:
fix has been released now (0.11.0). reopen if needed.

maun:
This is fixed, thanks.

ChanchalW:
![screen shot 2017-03-03 at 1 19 54 pm](https://cloud.githubusercontent.com/assets/21195828/23542611/3105f068-0014-11e7-8c67-7280e31c0b11.png)

Any suggestion? I was using this in another project in swift 2.2. Now i am using xcode 8.2 and facing this issue. Nothing helped about it.

raybradley:
Hi, I'm having this problem as of 11 March 2017 using Xcode 8.2.1, SQLite 0.11.2, and Carthage. Ideas?

nealsun:
same issue with @raybradley .

jberkel:
will be fixed once #615 is merged

phildow:
[...]

PR #502, opened 2016-09-17
Swift3

jberkel:
Supersedes #501, #475


jberkel:
@stephencelis it might make sense to skip #475, unless you want to keep a separate branch for 2.3 compatibility?

iLandes:
It's look great. 

How can I integrate it on a Pod file ?

jberkel:
@iLandes 

try
```ruby
pod 'SQLite.swift', :git => 'https://github.com/stephencelis/SQLite.swift.git',
       :branch => 'swift3-mariotaku'
```

cmvolare:
can you I get a carthage line to integrate this branch into my project? thanks in advance.

stephencelis:
Many thanks to @mariotaku for picking up my branch and getting it across the finish line! And many more thanks to @jberkel for doing an incredible job coordinating this migration. It's a big one!

stephencelis:
> it might make sense to skip #475, unless you want to keep a separate branch for 2.3 compatibility?
>
> Alternatively we could merge #475 first and then have a swift3 branch in parallel, but given the amount of changes this will become a maintenance nightmare.

It'd be nice to tag 2.3 compatibility, though I don't think maintaining both at once is feasible.

SQLite/Core/Blob.swift
```diff
@@ -30,16 +30,16 @@ public struct Blob {
         self.bytes = bytes
     }
 
-    public init(bytes: UnsafePointer<Void>, length: Int) {
-        self.init(bytes: [UInt8](UnsafeBufferPointer(
-            start: UnsafePointer(bytes), count: length
-        )))
+    public init(bytes: UnsafeRawPointer, length: Int) {
+        // TODO correct count
```
jberkel: @mariotaku what is needed here?

mariotaku: It works in tests and actual app, though I'm not sure whether my changes correct, maybe we could take a look at swift 2.3 branch.

CocoaPods/iphoneos-10.0/module.modulemap
```diff
@@ -0,0 +1,4 @@
+module CSQLite [system] {
+    header "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.0.sdk/usr/include/sqlite3.h"
```
sam-w: This hard-coding of the Xcode path causes build errors (Carthage, CocoaPods or just building in Xcode) for all users who don't have their copy of Xcode 8.x at that path.

CocoaPods/iphonesimulator-10.0/module.modulemap
```diff
@@ -0,0 +1,4 @@
+module CSQLite [system] {
+    header "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator10.0.sdk/usr/include/sqlite3.h"
```
sam-w: As above, this causes build errors for all users who don't have their copy of Xcode 8.x at that path.

CocoaPods/macosx-10.11/module.modulemap
```diff
@@ -0,0 +1,4 @@
+module CSQLite [system] {
+    header "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/usr/include/sqlite3.h"
```
sam-w: As above, this causes build errors for all users who don't have their copy of Xcode 8.x at that path.

CocoaPods/macosx-10.12/module.modulemap
```diff
@@ -0,0 +1,4 @@
+module CSQLite [system] {
+    header "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include/sqlite3.h"
```
sam-w: As above, this causes build errors for all users who don't have their copy of Xcode 8.x at that path.

CocoaPods/macosx/module.modulemap
```diff
@@ -1,4 +1,4 @@
 module CSQLite [system] {
-    header "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/usr/include/sqlite3.h"
+    header "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sqlite3.h"
```
sam-w: As above, this causes build errors for all users who don't have their copy of Xcode 8.x at that path.

CocoaPods/iphoneos-10.0/module.modulemap
jberkel: What do you suggest to do instead?

stephencelis: This hard-coding is required for CocoaPods. If you include SQLite.swift as a sub-project or via Carthage it uses an SDK-relative path instead.

I've raised this issue with CocoaPods in the past: https://github.com/CocoaPods/CocoaPods/issues/3942

We're always open to alternative solutions for CocoaPods users, though, and would happily take contributions that solve the problem more elegantly for them!

SQLite/Core/Blob.swift
jberkel: looks good to me, I added a bunch of tests to very the behaviour

jberkel:
PR  is ready for another round of review.

acrookston:
I was just upgrading my app and something isn't working for watchOS. Wanted to leave a note about it. Will be digging in deeper to understand why.

CocoaPods/iphoneos-10.0/module.modulemap
jberkel: @stephencelis this also affects Carthage builds – without the modulemap, the `CSQLite` imports cannot be resolved. Even worse, they're still required *after* the framework has been built (see #492).

jberkel:
Since I did not hear any more bug reports/complaints I just went ahead and merged the PR. The modulemap problems for non-CocoaPods users should hopefully be gone now.

PR #517, opened 2016-10-09
use sqlite3_trace_v2

jberkel:


iNoles:
we could do like 

```swift
#if swift(>=2.3)
// this compiles on Xcode 8 / Swift 2.3 / iOS 10
#else
// this compiles on Xcode 7 / Swift 2.2 / iOS 9
#endif
```

jberkel:
@iNoles this won't do but  gave me the idea to exclude osx via conditional compilation.

PR #553, opened 2016-12-05
Re-add support for SQLcipher via CocoaPods

jberkel:


stephencelis:
Rad! 👍

jberkel:
@stephencelis just released 0.11.1 with SQLiteCipher support. Maybe would be a good idea to add a deprecation note to https://github.com/stephencelis/SQLiteCipher.swift?

PR #548, opened 2016-11-19
Add Swift Package Manager support

wadetregaskis:
This is a slightly meandering series of patches that ultimately add support for the Swift Package Manager.  It's probably incomplete - though I did go to the effort of making the existing Xcode project still build after all these changes, I didn't test the tests nor the other package manager pieces (e.g. CocoaPods).

Though I haven't tested this anywhere but macOS 10.12, with a bit of luck this also works on Linux (re. issue #315).

PR #560, opened 2016-12-10
Swift Package Manager

jberkel:
Supersedes #548


jberkel:
@wadetregaskis is there a way to refer to the system CSQlite target instead of including the amalgamation?

Package.swift
```diff
@@ -0,0 +1,16 @@
...
+    targets: [
+        Target(
+            name: "SQLite",
+            dependencies: [
+                .Target(name: "SQLiteObjc")
+            ]),
+        Target(name: "SQLiteObjc")
+    ],
+    dependencies: [
+        .Package(url: "https://github.com/jberkel/CSQLite.git", majorVersion: 0)
```
jberkel: @stephencelis this should probably be moved to your namespace – I couldn't find a way to avoid a separate repository. there's something like dependency pinning now but it seems to be a newer feature, it's not on my system.

jberkel:
@stephencelis could you just check the one comment above – basically need you to clone the repo.

also, great stuff regarding today's Kickstarter OSS announcements, just bit disappointed that there's no sign of SQLite.swift in the iOS app :)

stephencelis:
@jberkel Thanks! Sorry for the delay. I've forked the dependent repo in case you wanted to make things consistent!

Yeah, the app is mostly online-only right now, so no need for a lib like SQLite.swift 😛, which might also show why my participation here has waned a bit—it's much harder to actively maintain libraries that you're not currently using. Happy holidays! Thanks again for being an invaluable contributor!

PR #591, opened 2017-01-16
Add possibility to have expression on right hand side of like

mcfedr:
Sometimes you need to use like with an expression on the right, i.e. the pattern is in the database

My use case is searching in a list of domain names, the table contains a single column domain, the SQL I want is:

    WHERE 'www.facebook.com' LIKE '%' || domain;

The code, with the changes is

    "www.facebook.com".like("%" + Expression<String>("domain"))


mcfedr:
@stephencelis any chance you want to merge this? or any changes you would like to see?

jberkel:
thanks, moved this PR into release branch

PR #597, opened 2017-02-06
Only import CSQLite module when building with CocoaPods

ceyhuno:
Issue #492 re-appeared so I made the change in db8f39b which was done in PR #502 again to fix it.

I'm using Xcode 8.2.1 and my Xcode app is placed in `/Applications/Xcode.app`. I only have one Xcode app installed.


jberkel:
dup of #615

yusufseyrek:
I strongly believe that #615 > #597 :) But thank you anyway @jberkel

jberkel:
@ceyhuno @yusufseyrek this is merged now, I'll soon push 0.11.3 if you want to use a released version.

PR #640, opened 2017-04-30
Allow transactions to be nonescapable

mohamede1945:


jberkel:
Sorry need more context. “Allow”? Isn't it actually more restricting?

mohamede1945:
Simply because the design of the library has everything synchronous. Insert/Delete/Update/Select all operations are synchronous, why a transaction that includes them will expect them to be asynchronous?

Secondly, a real-world example:

```swift
func saveAndGetUsers(userInfo: UserInfo) -> [User] {
    db.run(users.insert(userInfo))
    let rows = db.prepare(users)
    return rows.map { User($0) }
}
```

If I want to add them to be in a transaction, all I need to do is to wrap that into a non-escapable closure.

```swift
func saveAndGetUsers(userInfo: UserInfo) -> [User] {
    db.transaction {
        db.run(users.insert(userInfo))
        let rows = db.prepare(users)
        return rows.map { User($0) }
    }
}
```


I hope you see what I mean.

jberkel:
OK, had a closer look at the code. The closure does not need to be retained since it is executed synchronously anyway.

Couldn't the `sync` method just be simplified to:

```swift
func sync<T>(_ block: () throws -> T) rethrows -> T {
    if DispatchQueue.getSpecific(key: Connection.queueKey) == queueContext {
        return try block()
    } else {
        return try queue.sync(execute: block)
    }
}
```

Also not sure what that radar reference (`rdar://problem/21389236`)  is for, maybe a workaround? The number does not resolve to anything. @stephencelis ?

mohamede1945:
The radar was in your code already, I didn't add it. 

Regarding the simplification, I think DispatchQueue.sync accepts `() -> T` while the block parameter is of type `() throws -> T`.

jberkel:
There is a method with the exact method signature. Maybe it wasn't available when the code was written. Anyway I've applied your changes in the release branch, so closing this PR now. Would be great if you could try it, preferably with a larger code base.

mohamede1945:
Perfect. Thanks for incorporating the changes.

PR #649, opened 2017-05-08
Make Row.get throw instead of crash.

nickmshelley:
This isn't backwards compatible with the old `.get`, but subscripts still work the same. I figure most use subscripts so this won't be too big of a deal, but if you want me to think of a backwards compatible way to do this, let me know.


jberkel:
Applied in bd1fc80414c89b8d7b756899cff2a153d5fd04f1

PR #666, opened 2017-06-11
fix create/drop index functions

kireev165:
According to documentation I can do this:
```swift
try db.run(users.createIndex(email, ifNotExists: true))
// CREATE INDEX IF NOT EXISTS "index_users_on_email" ON "users" ("email")
```
and this:
```swift
try db.run(users.createIndex(email, ifNotExists: true))
// CREATE INDEX IF NOT EXISTS "index_users_on_email" ON "users" ("email")
```
But in fact:
```swift
import SQLite

let table = Table("table")
let field = Expression<Int64>("field")
table.createIndex(field, unique: true, ifNotExists: true)
// error: cannot convert value of type 'Expression<Int64>' to expected argument type '[Expressible]'
// table.createIndex(field, unique: true, ifNotExists: true)
table.dropIndex(field, ifExists: true)
// error: MyPlayground.playground:6:17: error: cannot convert value of type 'Expression<Int64>' to expected argument type '[Expressible]'
// table.dropIndex(field, ifExists: true)
```

So I removed:
```swift
public func createIndex(_ columns: Expressible...) -> String
public func dropIndex(_ columns: Expressible...) -> String
```
and changed:
```swift
public func createIndex(_ columns: [Expressible], unique: Bool = false, ifNotExists: Bool = false) -> String
public func dropIndex(_ columns: [Expressible], ifExists: Bool = false) -> String
```
to:
```swift
public func dropIndex(_ columns: Expressible..., ifExists: Bool = false) -> String
public func dropIndex(_ columns: Expressible..., ifExists: Bool = false) -> String
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,9 +109,7 @@ For a more comprehensive example, see [this article](http://masteringswift.blogs
 
 ## Installation
 
-> _Note:_ SQLite.swift requires Swift 3 (and [Xcode][] 8) or greater. If you absolutely
-> need compatibility with Swift 2.3 you can use the [swift-2.3][] branch or older
-> released versions. New development will happen exclusively on the master/Swift 3 branch.
+> _Note:_ SQLite.swift requires Swift 3 (and [Xcode][] 8). Use the [swift-4][] branch for Xcode 9 Beta.
 
 ### Carthage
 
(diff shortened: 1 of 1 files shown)


kireev165:
iOS version was updated to solve build error:
> xcodebuild: error: Unable to find a destination matching the provided destination specifier:
		{ platform:iOS Simulator, OS:10.3, name:iPhone 6s }
	The requested device could not be found because no available devices matched the request.

But now there is a new problem that I don't really understand.
So, anyone, please look at travis logs.

jberkel:
@kireev165 if you rebase against master it should work

kireev165:
@jberkel thanks. it seems that everything fine now

jberkel:
applied: ac52569861e6ced88f168be1fca2e20bcd1118cd

Issue #664, opened 2017-06-08
XCode 9.0 Beta : Compile Error


dinobei:
After updating to XCode 9 Beta, the following Compiler Error occurred (SQlite.swift 0.11.2). How can I solve it?
![2017-06-08 11 44 23](https://user-images.githubusercontent.com/7797499/26910131-edf2670c-4c3f-11e7-96ba-209cb88ea9e4.png)

RLovelett:
Fully acknowledge that I have no idea what I am doing. But given that it is complaining about files not found I figured it would be worth starting there.

So looking in the app bundles for Xcode 8.3.3 vs Xcode 9.0.0 I found this:

### Xcode Version 8.3.3 (8E3004b)

```
/Applications/Xcode.app/Contents/Developer/Platforms/AppleTVOS.platform/Developer/SDKs/AppleTVOS.sdk/usr/include/sys/_symbol_aliasing.h
/Applications/Xcode.app/Contents/Developer/Platforms/AppleTVSimulator.platform/Developer/SDKs/AppleTVSimulator.sdk/usr/include/sys/_symbol_aliasing.h
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/usr/include/sys/_symbol_aliasing.h
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk/usr/include/sys/_symbol_aliasing.h
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/_symbol_aliasing.h
/Applications/Xcode.app/Contents/Developer/Platforms/WatchOS.platform/Developer/SDKs/WatchOS.sdk/usr/include/sys/_symbol_aliasing.h
/Applications/Xcode.app/Contents/Developer/Platforms/WatchSimulator.platform/Developer/SDKs/WatchSimulator.sdk/usr/include/sys/_symbol_aliasing.h
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-migrator/sdks/AppleTVOS.sdk/usr/include/sys/_symbol_aliasing.h
[...]

PR #668, opened 2017-06-15
Swift 4 support

thebluepotato:
What changes:
 - Replaces `import CSQLite` with the new `import SQLite3` directly exposed to Swift code. 
 - Updates the project to Swift 4.0, with recommended warning and a slightly edited `map` syntax regarding tuples.
 - All the .modulemap files have been removed since they are no longer necessary.

The aim of this PR is to have code that compiles without issues with Swift 4/Xcode 9. All tests pass, except for those regarding SQLCipher, which haven't been tested. `pod lib lint` fails due to errors relating to SQLCipher. My guess is that there needs to be some tweaking on how SQLCipher is imported in Swift now that there's a native access to the SQLite API.


Sources/SQLite/Typed/Query.swift
```diff
@@ -494,8 +494,9 @@ extension QueryType {
             return nil
         }
 
-        return " ".join(clauses.join.map { type, query, condition in
-            " ".join([
+        return " ".join(clauses.join.map { arg in
+            let (type, query, condition) = arg
+            return " ".join([
```
thebluepotato: That's the issue in #664 : the map closure take one tuple argument instead of its components. That's new in Swift 4, see [SE-0110](https://github.com/apple/swift-evolution/blob/master/proposals/0110-distingish-single-tuple-arg.md).

jberkel:
Thanks for getting the ball rolling. Hope travis will release a new Xcode image soon.

thebluepotato:
So actually Travis succeeds, the only fails are in the validator schemes where there's some kind of timeout. The same commit has succeeded in Travis for my own branch : https://travis-ci.org/thebluepotato/SQLite.swift/builds/248984684.
I think we could merge into the swift 4 branch.

abunur:
I updated my Podfile as such:
pod 'SQLite.swift',  :git => 'https://github.com/stephencelis/SQLite.swift.git', :branch => 'swift-4'
but I am still getting that error 'sys/_symbol_aliasing.h' file not found

any thoughts?

thanks

thebluepotato:
@abunur Well that's because the `swift-4` branch hasn't been updated yet. That's why there's this pull request. We just need someone with write access to merge the PR.
Btw, I forgot to mention that the Travis builds now work because there _is_ an Xcode 9 image now.

abunur:
@thebluepotato lol my bad sorry :)

abunur:
@thebluepotato so after switching to your branch, doing a pod install --repo-update
and an option-clean in Xcode, I am still getting the same error.  Any other steps I need to do?  Running Xcode 9 Beta 2.

thebluepotato:
@abunur I've tried switching to the latest release, `pod update`, build (the error showed up) then switching back to my branch, `pod update` and build (no error).
I'd suggest making sure that your Podfile is correct, saved before `pod update` and if the error still shows up, deintegrate Cocoapods using `pod deintegrate` or `pod superdeintegrate` and finally running `pod install`again.

jberkel:
@thebluepotato thanks! looks like there aren't any massive changes in swift4. and the simulator test instability bug is back...

thebluepotato:
@jberkel Indeed, when I closed and reopened the PR yesterday to trigger a rebuild, Travis worked completely but merging the most recent commit from `master`triggered another build which did fail here on the PR but worked on my branch...
Indeed, Swift 4 introduces mostly incremental changes. It's designed to be _mostly_ source code compatible with Swift 3. The new stricter difference between single-tuple and multiple arguments is one of the changes that needed some adjustment.

PR #690, opened 2017-07-21
Fix OS X database path

avinassh:
Currently, the `path` is set to: 

```
/Users/avi/Library/Application SupportMyIdentifier
```

This PR fixes it and adds the missing slash:

```
/Users/avi/Library/Application Support/MyIdentifier
```


jberkel:
thanks, part of 0.11.4

PR #699, opened 2017-08-03
try was missing before db.scalar

igordeoliveirasa:
try was missing before db.scalar


jberkel:
Thanks, part of 0.11.4

PR #703, opened 2017-08-17
Fixed all warnings and build errors

ezfe:
- Changed the `OBJC_INFERENCE` values to `Default`.
- Removed redundant constraints in `Sources/SQLite/Typed/Operators.swift`
- Switched from `String.subscript(to: Range)` to `String[..<Range]` in `Sources/SQLite/Extensions/FTS4.swift`
- Unwrap upper and lower utf8 bounds in `Sources/SQLite/Extensions/FTS4.swift`

Note that this includes the change made in #694, however I implemented it slightly different to be cleaner and more clear.


jberkel:
Sorry, somehow missed this PR, and duplicated your work. thanks anyway!

PR #722, opened 2017-09-07
add support for ORDER and LIMIT on UPDATE and DELETE

### LINKED ISSUE #657: LIMIT OFFSET in delete()
How to enable SQLITE_ENABLE_UPDATE_DELETE_LIMIT, because this one
```swift
let messages = Table("messages")
.filter(Expression<String>("chatId") == chat[Expression<String>("id")])
.order(Expression<Date>("createdAt").desc)
.limit(Int.max, offset: index > 29 ? 1 : 30)

try db.run(messages.delete())
```
removes ALL rows from `messages`

Dridus:
SQLite supports this syntax when compiled with the `SQLITE_ENABLE_UPDATE_DELETE_LIMIT` compile option, which Apple apparently does on at least iOS 10.3 and macOS 10.12. Even if it isn't compiled with this option, if the library user specifies an order and limit it gets silently ignored as-is, which seems dangerous.

this fixes #657

### DIFF
--- a/Sources/SQLite/Typed/Query.swift
+++ b/Sources/SQLite/Typed/Query.swift
@@ -648,7 +648,9 @@ extension QueryType {
             tableName(),
             Expression<Void>(literal: "SET"),
             ", ".join(values.map { " = ".join([$0.column, $0.value]) }),
-            whereClause
+            whereClause,
+            orderClause,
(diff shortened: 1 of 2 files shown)


jberkel:
Thanks, part of the 0.11.4 branch now

PR #723, opened 2017-09-07
Added support for the union query clause

stephanheilner:


jberkel:
thanks, this is part of of the 0.11.4 release branch

PR #724, opened 2017-09-13
Fix compiler warnings

quaddef:


jberkel:
Which compiler warnings?

PR #726, opened 2017-09-17
Add cursor type for more safety.

nickmshelley:


Sources/SQLite/Typed/Query.swift
```diff
@@ -894,58 +894,89 @@ public struct Delete : ExpressionType {
...
+                            try expandGlob(true)(q)
+                            continue column
                         }
                         throw QueryError.noSuchTable(name: namespace)
                     }
-                    for q in queries {
-                        try expandGlob(query.clauses.join.count > 0)(q)
-                    }
-                    continue
+                    fatalError("no such table: \(namespace)")
```
jberkel: this should probably be `QueryError.noSuchTable`?

nickmshelley: Good call. Sorry I took so long to make this simple change.

jberkel:
I merged this into the release branch, with some additional changes (5a9b583d2a28d12bc8523b204e75a0da6c8b04ff). I'm wondering if `RowCursor` will be confused with `Cursor` on statement, maybe a different name would be more appropriate? `RowIterator`?

nickmshelley:
Yeah `RowIterator` sounds fine to me. Thanks for getting it in!

Issue #727, opened 2017-09-19
SQLiteEncoder and SQLiteDecoder for Swift 4


drewag:
I am working on creating a `SQLiteEncoder` and a `SQLiteDecoder` so that I can insert and retrieve objects from the database by simply implementing the `Codable` and `Decodable` protocols introduced in Swift 4. I originally implemented this outside of the library with my own Coding protocols and I am in the process of converting it to use Swift 4 Coding protocols instead of my custom ones. I was originally going to leave this outside of the framework but I discovered I will need internal access to some of the objects.

**The Plan**
- I do not plan to support relationships between objects or anything like that.
- The SQLiteEncoder will essentially just generate a list of setters for the encodable object. I will probably add a convenience insert method like `insert<V: Encodable>(_ encodable: V) -> SQLite.Insert` to QueryType that accepts an Encodable type and calls insert with the generated setters
- SQLiteDecoder will just do a series of gets on a provided row. I will probably add a method to `Row` like `decode<V: Decodable>() throw -> V`.
- The coders will simply convert all non-native types to and from JSON data

Is anyone already attempting this or think that this is an undesirable feature inside this framework?

PR #728, opened 2017-09-20
Add Coding support

drewag:
- Insert or update from an Encodable type
- Decode a row into a Decodable type

Closes stephencelis/SQLite.swift#727


jberkel:
Thanks for the PR. could you rebase it against the release branch?

jberkel:
→ #733

PR #749, opened 2017-10-17
Do not use .characters on Strings

dfed:
Prevents warnings in Xcode 9.1 by adopting proper Swift 4 syntax.

cc @noahsmartin @fdiaz


jberkel:
thanks!

PR #753, opened 2017-11-07
Resolved the warning that was raised because "characters" of a "String" has been deprecated.

vinayganesh:
Resolved a warning that was caused dude to "characters" deprecation in "String"


martheli:
When I use  pod 'SQLite.swift', '~> 0.11.4', the master branch with this change is not included.

PR #774, opened 2018-01-16
Add Array Unterface for createIndex(_:unique:ifNotExists:)

h-yuya:
How about adding array interface for createIndex(_:unique:ifNotExists:), in addition to variadic interface?
Inside variadic function the type of the parameter is Array. But it does not accept Array Type.
Isn't it inconvenient a little bit in using it inside some functions like below?

```swift
import SQLite

var table: Table
func Hoge(numberColumns: Int...) {
...
table.createIndex(numberColumns) // Error because of variadic.
...
}
```

Sincerely,

### DIFF
--- a/Sources/SQLite/Typed/Schema.swift
+++ b/Sources/SQLite/Typed/Schema.swift
@@ -126,16 +126,20 @@ extension Table {
     }
 
     // MARK: - CREATE INDEX
-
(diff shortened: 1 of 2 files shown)


kireev165:
It is inconvenient and it’s my fault. I tried to remove controversy between documentation and code.
Please see https://github.com/stephencelis/SQLite.swift/blob/master/Documentation/Index.md#indexes and https://github.com/stephencelis/SQLite.swift/pull/666 for more details.

PR #799, opened 2018-04-16
Add `and` and `or` functions

davedelong:
When you're programmatically building a query to execute, you have times where you build a variable number of clauses by which you filter a table. There isn't a great way to turn an `Array<Expression<Bool>>` into an `Expression<Bool>`, short of making a large `(a AND (b AND (c AND (d AND e))))` expression.

`and` and `or` fix this by allowing you to turn an `Array<Expression<Bool>>` directly into a lower-complexity clause: `(a AND b AND c AND d AND e)`

### DIFF
--- a/Sources/SQLite/Helpers.swift
+++ b/Sources/SQLite/Helpers.swift
@@ -73,7 +73,11 @@ extension String {
     }
 
     func infix<T>(_ lhs: Expressible, _ rhs: Expressible, wrap: Bool = true) -> Expression<T> {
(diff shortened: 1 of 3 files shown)


KyleLeneau:
Cool!

stephanheilner:
Would this replace the `.filter`?

davedelong:
A year and a half later, I find myself needing this again.

Is there anyone around who's interested in merging this?

NathanFallet (approved):

PR #801, opened 2018-04-21
Add an optional target queue to for database operations

sharplet:
The optional target queue is useful for synchronising other work with respect to database operations. Provided the target queue is also a serial queue, work submitted to the private connection queue and work submitted directly to the target queue will still be executed in serial.

The queue context is now set on the target queue if present. The connection queue inherits values set on the target queue, so the call to `DispatchQueue.getSpecific`, using `queueKey`, returns `queueContext` as expected. For this logic to still be thread-safe, the target queue must be a serial queue. `queueKey` is now an instance variable so that every connection gets a unique key, ensuring that connections can't write over one another's context variables.

The `attributes:` parameter is also removed from the call to initialise the database queue, as `[]` (i.e., serial) is already the default.


Sources/SQLite/Core/Connection.swift
```diff
@@ -119,11 +129,18 @@ public final class Connection {
     ///
     ///     Default: `false`.
     ///
+    ///   - target: An optional target queue for database operations to be executed on.
+    ///     Set this to a custom serial queue if you need to schedule work that must be
+    ///     synchronized with respect to database operations.
+    ///
+    ///     **Important:** The target queue **must not** be a concurrent queue, or access
+    ///     to the returned connection is no longer guaranteed be thread-safe.
```
jberkel, line 137: could this be asserted in code?

Sources/SQLite/Core/Connection.swift
```diff
@@ -101,11 +101,21 @@ public final class Connection {
...
+    ///
+    /// - Throws: `Result.Error` iff a connection cannot be established.
+    ///
     /// - Returns: A new database connection.
-    public init(_ location: Location = .inMemory, readonly: Bool = false) throws {
+    public init(_ location: Location = .inMemory, readonly: Bool = false, target: DispatchQueue? = nil) throws {
+        queue = DispatchQueue(label: "SQLite.Database", target: target)
```
jberkel, line 115: looks like a queue is created here even though it might not be needed (when `target` is set)?

Sources/SQLite/Core/Connection.swift
sharplet: That would be nice! Currently I'm not aware of public API to check whether a queue is serial or concurrent. 🤔

Sources/SQLite/Core/Connection.swift
sharplet: Targeted queue hierarchies are meant to be very cheap, so this is intentional.

PR #804, opened 2018-04-25
Removed unnecessary header import, fixes build process when using cocoapods and static libraries.

stefanceriu:
SQLite.h bridging header import fails when using static libraries with cocoapods 1.5.0

`...Pods/SQLite.swift/Sources/SQLite/SQLite.h:6:9: 'SQLite/SQLite-Bridging.h' file not found`

**STRs**
Use Cocoapods 1.5.0

```
source 'https://github.com/CocoaPods/Specs.git'

platform :ios, '9.0'

target 'TestApplication' do

  pod 'SQLite.swift'

end
```


markhim:
#import "SQLite-Bridging.h"

stefanceriu:
If that works with cocoapods static libs then sure.. ¯\_(ツ)_/¯

PR #832, opened 2018-08-06
Fix for non-framework build

Udumft:
Fixed a mistype that prevented building SQLite.swift as non-framework sources.

Steps to reproduce:
- download the source
- create new XCode project
- add SQLite.swift sources inside new project (just sources, not the actual .xcodeproj file)
- build.

As result: sources are built as is and no framework is created -> reference to `SQLite.` becomes invalid. 
Keeping in mind that `Setter` is referenced in this file earlier I assume this is a mistype.


jberkel:
thanks!

PR #843, opened 2018-09-30
added sample code for copying a 'seed' database

jasonpeterson:


jasonpeterson:
Should I try this again? Not sure why a docs change would fail one of the 11 builds, but the error is vague: "unexpected service error: The Xcode build system has crashed."

Issue #652, opened 2017-05-12
Limited number of PRIMARY KEY constraints


Phynixx:
I want to add 4 PRIMARY KEY constraints, but there are just functions to add max. 3, although all of them call the following function, which accepts a variable number of keys:
```
fileprivate func primaryKey(_ composite: [Expressible]) {
        definitions.append("PRIMARY KEY".prefix(composite))
    }
```
So could you please make this function public?

PR #853, opened 2018-10-16
add primary key with 4 columns

hellohuanlin:
Hi, I need to support 4 columns as the compound key. I am not able to create an extension because `definitions` variable `primaryKey` helper function are both `fileprivate`. 

My current temp solution is to write the raw query, but i think it's better to update the library. 

4 should be enough for most of projects. 

related issue: https://github.com/stephencelis/SQLite.swift/issues/652


hellohuanlin:
cc @jberkel really appreciate your maintenance work on this library :). Any chance you could take a look 🙏?

ypopovych:
Hello @hellohuanlin. Can you rebase it to the latest master? I think this is good PR to merge

PR #864, opened 2018-12-06
updated Bridging header to be compatible with SQLCipher 4.0.0 pod

asidden:
until this change, I was not able to install SQLite.swift with Xcode 10.1 via
``` pod 'SQLite.swift/SQLCipher'

SQLCipher dependency pod version resolved - 4.0.0

the build failed with error:

error: include of non-modular header inside framework module 'SQLite.SQLite_Bridging': '<project-path>/Pods/SQLCipher/sqlite3.h'
#import "sqlite3.h"

however, I'm not sure this change does not affect Carthage or SPM installation.

### DIFF
--- a/Sources/SQLiteObjc/include/SQLite-Bridging.h
+++ b/Sources/SQLiteObjc/include/SQLite-Bridging.h
@@ -23,8 +23,7 @@
 //
 
 @import Foundation;
-
-#import "sqlite3.h"
+@import SQLCipher;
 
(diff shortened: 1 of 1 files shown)


jonathanrfisher1:
This resolved my Xcode 10 issue with the same error.  Looks like it's breaking CI though: module 'SQLCipher' not found

Do you know a fix for this @asidden ?

For now I plan on using your change with a git patch.

Issue #777, opened 2018-01-23
Select from multiple tables


JanC:
## Build Information

- Include the SQLite.swift version, commit or branch experiencing the issue.
0.11.4
- Mention Xcode and OS X versions affected.
Xcode 9.2,  iOS 11
- How do do you integrate SQLite.swift in your project?
    - Carthage

Hi,
I noticed that when selecting all columns from multiple tables, SQLite.swift throws an `No such table:` exception even if the table does exists and the generated SQL itself works fine

More details are in the [SO question](https://stackoverflow.com/questions/48404559/sql-swift-select-from-multiple-tables/)

Basically
```swift
users.select(users[*], posts[*]).join(posts, on: user_id == users[id])
```
would throw "No such table: posts"

while this works

```swift
users.select(users[*], posts[post_text]).join(posts, on: user_id == users[id])
```

In both cases, the generated SQL is correct

cheers,
Jan

jsi7iv:
Having the same problem. 

```
actionAlias.select(modelTableAlias[*], actionAlias[*])
			.join(JoinType.leftOuter, modelTableAlias, on: actionAlias[ActionDao.modelId] == cast(modelTableAlias[idColumn]) as Expression<String> )
            .filter( actionAlias[ActionDao.modelTable] == tableName)
```

This worked just fine before.

nguyenthinam:
I got the same problem, please fix asap.

azmikemm:
Proposed  Solution for this issue  #777
I had the same problem, and debugged the SQLite.swift code.  In file "Query.swift" in function  "columnNamesForQuery(_ query: QueryType)" in the code block following "if column == "*" {" in the loop "for q in queries {" there is a "throw QueryError.noSuchTable(name: namespace)" immediately after the if statement **WITHIN** the loop.  Therefore the loop tests one entry (first table of the join), then always throws.  The fix is to delete this mis-placed throw statement, leaving the second throw (outside the loop) in place. (Sorry I can't seem to get indenting to work on the copied code below).

if column == "*" {
                var select = query
                select.clauses.select = (false, [Expression<Void>(literal: "*") as Expressible])
                let queries = [select] + query.clauses.join.map { $0.query }
                if !namespace.isEmpty {
                    for q in queries {
                        if q.tableName().expression.template == namespace {
                            try expandGlob(true)(q)
                            continue column
                        }
                        **throw QueryError.noSuchTable(name: namespace) <- DELETE THIS LINE!**
[...]

PR #873, opened 2019-01-16
attempt to fix issues with join statement

asidden:
there is an error thrown, when query is built with join and select is set to specific columns from joined table (see https://github.com/stephencelis/SQLite.swift/issues/777 and others)

applied a possible fix + added a test to a test-suit


asidden:
closing this pr, as it is not up-to date with the latest master
see updated pr at https://github.com/stephencelis/SQLite.swift/pull/916

PR #876, opened 2019-01-30
CocoaPods static libraries support

MrAsterisco:
I have tried to include this library as a CocoaPods dependency in a iOS application where the `use_frameworks!` option is not enabled. The reason to remove this option is mainly related to startup times, as reported by multiple sources around the web.

The issue with this library is related to how the `SQLite-Bridging.h` is linked in `SQLite.h`. When included as a static library, the aforementioned file can't be reference using `<>` but needs to be imported as a normal file.

I've made the necessary changes and I'd like to see these implemented in the main project. I'm not aware of incompatibilities with other package managers (such as Carthage).


danramteke:
This change works for me.

MrAsterisco:
@danramteke Thank you! How do we proceed? I don't have permissions to merge this pull request.

ypopovych:
Do we have this problem with the latest one?

danramteke:
@MrAsterisco I am not an admin on this repo.

isnine:
This change works for me too.

PR #881, opened 2019-02-22
Add support for custom SQL aggregates

kasei:
Would there be interest in supporting SQLite custom aggregates (and if so, possibly also custom window functions)? This PR adds support for custom aggregates implemented in swift, allowing both a relatively low-level API requiring unsafe pointers to maintain state, as well as a more user-friendly, `reduce`-style API. I've attempted to follow the style of the existing code implementing custom SQL functions, and added tests for both API styles.


NathanFallet (approved):

jberkel:
This wasn't quite ready for merging yet. The tests crash intermittently, investigating.

jberkel:
> Do not assign an instance of a nontrivial type through pointee to uninitialized memory. Instead, use an initializing method, such as initialize(to:count:).

6d5bb0ca1d561589cbcf5ffbacd154e5c05659cf

Issue #842, opened 2018-09-25
Support UPSERT using update in addition to replace


moshegottlieb:
Currently, the ONCONFLICT support for UPSERT is `replace`.  
While this works in most cases - it's not very useful when using cascading foreign keys because `replace` means the old record is **deleted** before a new record is added, causing dependent records via foreign keys to be deleted as well.  
For instance:

```
// Turn on foreign keys
try connection.execute("PRAGMA foreign_keys = ON;")
// Create table user
let id = Expression<String>("id")
let name = Expression<String>("name")
try connection.run(Table("user").create(){
    t in
        t.column(id, primaryKey: true)
        t.column(name)
    })

// create table message
let user_id_ref = Expression<String>("id")
let id = Expression<String>("id")
let user_id = Expression<String>("user_id")
let body = Expression<String>("body")
try connection.run(Table("message").create(){
    t in
        t.column(id, primaryKey: true)
        t.column(user_id)
        t.column(body)
        t.foreignKey(user_id, references:Table("user"), user_id_ref,delete:.cascade)
    })
```

If you add a user, and then messages, and then do something like:
`try connection.run(table.insert(or: .replace, ....` - all the existing messages will be deleted for the inserted user.  
Instead, a new `onConflict` support should be added for `.update`, as discussed [here](https://www.sqlite.org/lang_UPSERT.html).

jberkel:
Added in #882

Issue #482, opened 2016-08-01
Is there any close connection method?


liangquandeng:

jberkel:
Currently not – the connection is automatically closed when the connection object is deallocated.

PR #882, opened 2019-02-24
Implement Upsert [#482]

andyfinnell:
# Problem

Apps need the ability to insert or, if the row already exists, update the existing row. This is commonly called "upsert" and is [documented for SQLite here](https://www.sqlite.org/lang_UPSERT.html). The existing "replace" on conflict clause for `insert()` often does not have the desired behavior. It will cascade (with deletes, if specified) along foreign keys if the value exists and has to be replaced ([see Issue 842](https://github.com/stephencelis/SQLite.swift/issues/842)).

# Solution

Create an `upsert()` method for queries. The basic form allows for the fallback setters to be manually specified. A slightly more ergonomic version will compute the setters from the inserted values by removing the "conflicting" column, and then using the inserted values via the "excluded" qualifier. Additionally, an `Encodable` version of `upsert` is provided.

### DIFF
--- a/Documentation/Index.md
+++ b/Documentation/Index.md
@@ -35,6 +35,7 @@
       - [Sorting Rows](#sorting-rows)
       - [Limiting and Paging Results](#limiting-and-paging-results)
       - [Aggregation](#aggregation)
+  - [Upserting Rows](#upserting-rows)
   - [Updating Rows](#updating-rows)
   - [Deleting Rows](#deleting-rows)
   - [Transactions and Savepoints](#transactions-and-savepoints)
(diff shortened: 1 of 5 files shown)


andyfinnell:
Travis is failing on the macOS builds because it is using High Sierra instead of Mojave. UPSERT syntax was added to SQLite with version 3.24.0 (2018-06-04). However, macOS apparently didn't include that version until Mojave.

nferruzzi:
since the main repo looks dead, did you by any change keep an updated fork with this patch?

andyfinnell:
I don't, sorry.

nferruzzi:
@andyfinnell at the end I found another fork with this PR and others merged, so I switched there; thank you for your work :)

https://github.com/thomasrzhao/SQLite.swift

PR #898, opened 2019-04-01
Update SwiftPM manifest file

Jeehut:
This adds support for SwiftPM manifest based dependency managers. Specifically this adds support for installing via [Accio](https://github.com/JamitLabs/Accio) but will probably also work with SwiftPM once it's integrated into Xcode.


jberkel:
It looks like Accio has been discontinued in the meantime. Reopen if it's still needed.

Issue #888, opened 2019-03-19
Invalid SQL statements with Xcode 10.2


tanzolone:
## Build Information

- SQLite iOS Version 0.11.4 (however it's likely to impact all the targets)
- Xcode 10.2 beta 4
- manual integration in our project

## General guidelines

In Xcode 10.2 beta 4 the macro` #function `shows a different behaviour. As SQLite.swift relies on `#function` in different subroutines for the composition of SQLite queries, several bugs linked to invalid SQLite statements seem to be introduced. 

I have just run the SQLite.swift tests with Xcode 10.2 beta 4 and **there are several failures linked to this issue**.

An example as follows:

In the following helper function (Helpers.swift line 107):

```
func wrap<T>(_ expression: Expressible, function: String = #function) -> Expression<T> {
    return function.wrap(expression)
}
```

when called from

`static func count(_ star: Star) -> Expression<UnderlyingType>`

the value of `function` when running in Xcode 10.2 beta 4 is `count(_:)` instead of `count`. This leads to invalid SQLite statements of the type `SELECT count(_:)(*) myTable`.

christopheblin:
I have the exact same problem with Xcode 10.2 (not the beta) !

Do you have a workaround or a fix ?

tanzolone:
The best option in my opinion is to get rid altogether of the `#function` default parameter and explicitly passing the name of the SQL function (maybe an enum with string raw value). The `#function` macro seems to be intended for debug purposes (this explains this sort of change in a minor update of Xcode) and I would avoid relying on it for such a critical part of the SQL statement composition.

christopheblin:
Note that this also silently breaks SqlLiteMigrationManager 

https://github.com/garriguv/SQLiteMigrationManager.swift/blob/a716a877eca2e25fd8f0201605feaebfc0db27fc/Sources/SQLiteMigrationManager.swift#L68

=> Will always fall into the catch and report that no migration exists ...

christopheblin:
Here is a patch to apply if you have a pod install

[mystuff.patch.zip](https://github.com/stephencelis/SQLite.swift/files/3008304/mystuff.patch.zip)

chess92:
the simplest fix would be to just ignore any string after the first `(` of `#function`

arjavlad:
> Here is a patch to apply if you have a pod install
> 
> [mystuff.patch.zip](https://github.com/stephencelis/SQLite.swift/files/3008304/mystuff.patch.zip)

[...]

PR #903, opened 2019-04-04
Fix SQL statement generation problem on Xcode 10.2

codepgq:
Referenced SQLite.swift issue  **[#888](https://github.com/stephencelis/SQLite.swift/issues/888)**  
fix check a row value problem


ypopovych:
It was already fixed. Check master branch, please. We fully removed `#function` approach

PR #902, opened 2019-04-03
Version 0.11.6

sburlewapg:
Per recent discussion, this bumps the version to 0.11.6 and updates the project to Swift 5.

First time contributing, so please forgive anything I've missed in this.

One change in Foundation.swift I'm not 100% sure about because it required force unwrapping `pointer.baseAddress` within `withUnsafeBytes`, but I wasn't able to find anything less invasive to the current implementation.

PR #911, opened 2019-04-18
Remove warnings about 'public' modifier is redundant

Devaria:
When compiling SQLite with the newest versions of Xcode 14 warnings are generated stating that:
```
'public' modifier is redundant for instance method declared in a public extension
```

This PR fixes that warning.

Alternatively, we could remove the `public` modifier from the 14 function declarations below, but as the extension contains both `public` and `fileprivate` functions I think it is more appropriate to remove the top-level public modifier on the extension itself. 

I think this was implemented in [SR-8453](https://bugs.swift.org/browse/SR-8453) and became a warning starting with Xcode 10.2, but this should not break compilation for any version of Xcode capable of compiling code for Swift 4.2 that this project is currently using.

### DIFF
--- a/Sources/SQLite/Typed/CustomFunctions.swift
+++ b/Sources/SQLite/Typed/CustomFunctions.swift
@@ -22,7 +22,7 @@
 // THE SOFTWARE.
 //
 
-public extension Connection {
+extension Connection {
 
     /// Creates or redefines a custom SQL function.
     ///


sburlewapg:
@Devaria has been fixed by #902 which is available in master.

ypopovych:
@sburlewapg @Devaria it depends on the number of methods. If you have 10 public and one fileprivate/internal/private - it's logical to make extension public, and change type only for this methods. It's simple syntactic sugar.

ypopovych:
I think it can be even split into 2 extensions, one only with public methods, and other with internal/private ones

Devaria:
While I would have gone with an internal protocol I don't think any of the suggestions here offers enough of an advantage over #902 to warrant an extended discussion. I am just happy the warnings are gone :)

PR #811, opened 2018-05-31
Fixed bug not allowing columns from multiple tables in the .select

stephanheilner:

PR #916, opened 2019-04-24
attempt to fix issues with join statement

asidden:
there is an error thrown, when query is built with join and select is set to specific columns from joined table (see #777 and others)

applied a possible fix + added a test to a test-suit

this is the same pr as (https://github.com/stephencelis/SQLite.swift/pull/873), but rebased on the master version with swift5 support


NathanFallet:
Should be fixed by #811. Tell me if there is any problem with that.

PR #919, opened 2019-04-26
Support for database backup (sqlite3_backup_init)

kkapitan:
Hi, 

First of all - thank you for this amazing utility! My team and I are using this in some parts of our application and we are planning to expand the usage even further. However one feature of sqlite that we rely on is the ability to dump in memory databases to a file. It would be great to have that as a part of this library and that's why I'm leaving this Pull Request here with an example implementation. 

I'm happy to hear any feedback from you and thanks again for your work!


jberkel:
@kkapitan  thanks for the PR, looking good. I think the API would be more intuitive if the source parameters came first:

```swift
return try Backup(sourceConnection: self,
                  sourceName: databaseName,
                  targetConnection: targetConnection,
                  targetName: targetDatabaseName)
```

Issue #920, opened 2019-04-26
'UnsafeRawBufferPointer' is not convertible to 'UnsafePointer<_>'


Viswanth92:
I used this integration to report this error.

use xcode 10.1 ,swift 4.2

I tried pod version 0.11.6 and the issue still exists. I do not want to update my Xcode yet due to other SDKs that I am using my project.

aniiskywalker:
Any solutions?

bdorfman:
Having the same issue, it seems that  0.11.6+ should be marked as "Swift 5 Required" and not "Swift 5 Supported"

It seems like possibly replacing the implementation of `datatypeValue` in `Foundation.swift` with the following could be a possible solution to continue supporting both versions but I am unclear how safe this change actually is.

```    public var datatypeValue: Blob {
        #if swift(>=5.0)
        return withUnsafeBytes { (pointer: UnsafeRawBufferPointer) -> Blob in
            return Blob(bytes: pointer.baseAddress!, length: count)
        }
        #else
        return withUnsafeBytes { (pointer: UnsafePointer<UInt8>) -> Blob in
            return Blob(bytes: pointer, length: count)
        }
        #endif
    }```

If you make this change and also change SWIFT_VERSION to 4.2, all the tests seem to pass on Xcode 10.1 at least

JasonLiu1989:
try this
add these code into your podfile
`post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      config.build_settings['SWIFT_VERSION'] = '4.2'
      config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '9.0'
    end
  end
end`
and set pod 'SQLite.swift', '0.11.5'

PR #930, opened 2019-05-20
If not on Swift 5+, use old datatypeValue implementation

bdorfman:
This fixes compiler errors when using the latest release of SQLite with Swift 4.2/Xcode 10.1 (as discussed in this issue: https://github.com/stephencelis/SQLite.swift/issues/920)

I notice the test suite is set up to only build and run with SWIFT_VERSION=5.0 To verify this fix, I manually changed the version to 4.2 and ran it in Xcode 10.1. However it might be desirable to also duplicate all test targets and have them also run on 4.2 if continued support of the previous version is desirable.

### DIFF
--- a/Sources/SQLite/Foundation.swift
+++ b/Sources/SQLite/Foundation.swift
@@ -35,9 +35,15 @@ extension Data : Value {
     }
 
     public var datatypeValue: Blob {
+        #if swift(>=5.0)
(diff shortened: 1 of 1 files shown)


NathanFallet:
@jberkel Is it planned to keep the support for Swift 4.2 and lower?

jberkel:
I would prefer to move on to Swift 5.x.

NathanFallet:
We dropped the support of Swift 4.2 in `0.13.0`.

PR #951, opened 2019-08-16
Codable insert support onConflict

johannesebeling:
Simple addition to the codable api to pass a onConflict parameter to the insert function.

### DIFF
--- a/Sources/SQLite/Typed/Coding.swift
+++ b/Sources/SQLite/Typed/Coding.swift
@@ -44,6 +44,30 @@ extension QueryType {
         try encodable.encode(to: encoder)
         return self.insert(encoder.setters + otherSetters)
     }
+    
+    /// Creates an `INSERT` statement by encoding the given object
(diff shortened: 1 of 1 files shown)


OzzyTheGiant:
I'd like this merged in too. I have a project that needs to save batches of Codable objects by either inserting or updating during a transaction. Additionally, it would be nice if while inserting Codables, the library can transform an id of zero into the next available number in an autoincremented primary key, if this is not available already; I say that because my Codable objects have a non optional Int id so the default has to be 0 for brand new objects.

johannesebeling:
@NathanFallet Thank you! I've been waiting for this day for so long. This was one of my first legit open source contributions and seeing it finally merged feels great!

NathanFallet:
@johannesebeling Thanks for your contribution! This repository was inactive for a while, but I joined the maintainers team a few days ago, and I'm trying to review and merge pull requests, and respond to issues 😄 (it's a lot of work)

PR #971, opened 2019-11-25
Bump watchOS deployment version to 3.0

dchohfi:
Thank you for the library, we need to bump it to version 3.0 for watchOS to be able to submit to Apple using Carthage.


jberkel:
already targeting 3.0, closing

PR #976, opened 2019-12-26
Swift 5.1 Property Wrapper:  Expressed

ytyubox:
Consider of [SE-0258](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md)

```swift
@Expressed("id") var id:String = UUID().string
```
## Help Wanted 
1. `Expressed` Maybe not a proper name, help wanted.
2. Still working on tests, don't know how to test for a property wrapper, Equable?
3. Edit README and Documentation.

### DIFF
--- a/Sources/SQLite/Typed/Expressed.swift
+++ b/Sources/SQLite/Typed/Expressed.swift
@@ -1,13 +1,4 @@
-//
-//  File.swift
-//  
-//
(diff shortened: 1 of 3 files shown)


jberkel:
I think this is a good idea but this needs more work. I'm also not sure we need the extra `insert` methods.

PR #994, opened 2020-04-15
fix namespace check when joining tables

balland:
The unit test is reproducing the issue. Even if the query is correct, a
"table not found error" was returned.

This PR is based on https://github.com/stephencelis/SQLite.swift/pull/916

Fix https://github.com/stephencelis/SQLite.swift/issues/777


NathanFallet:
Should be fixed by #811. Tell me if there is any problem with that.

PR #1002, opened 2020-05-14
Always start the iterator at the same location

gsabran:
I was looking at a crash that happens when `statement.next()` is called. I have not gotten to the bottom of it, but I came across this code that doesn't seem to do what it intends. Indeed it seems that calling the iterator several time will not do the same thing:
```swift
let result = self.db.prepare(query)
result.forEach { ... }
// The second call will loop over the rows from a different element?
result.forEach { ... }
```


gsabran:
Next is actually called from within iterator, not when creating it. I got confused.

PR #1024, opened 2020-10-02
Change the behavior, of the sqlite3 import if it has SQLCipher as a d…

Domivat:
Change the behavior, of the sqlite3 import if it has SQLCipher as a dependency


khsora34:
This fixes the
`Redefinition of 'sqlite3_pcache_methods2' `
and similar issues on Xcode

jberkel:
applied in 3476f89153bcd31cc2836bdcc6ae95875ed6eb1c

PR #1027, opened 2020-10-08
Add a new method to capture errors during the iteration

gsabran:
The iterator returned by `prepare` would crash during iteration when it encounters an error at

https://github.com/stephencelis/SQLite.swift/blob/210098546d59678f720e4ef67f8b562acfebbdb0/Sources/SQLite/Core/Statement.swift#L211

This PR proposes a change to add a method to load all of the query's results at once, safely catching possible errors.

### DIFF
--- a/Sources/SQLite/Typed/Query.swift
+++ b/Sources/SQLite/Typed/Query.swift
@@ -917,6 +917,8 @@ public struct RowIterator: FailableIterator {
 
 extension Connection {
 
+    /// Return an iterator that will lazily load the query's results as they are iterated over.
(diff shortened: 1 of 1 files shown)


NathanFallet:
@gsabran I merged #1030, does it fix the problem or this PR needs to be merged as well? (because it is related)

jberkel:
the suggested chance is problematic for large queries (memory)

jberkel:
actually this is already possible with the `RowIterator` API:

```swift
let result = try Array(db.prepareRowIterator(users))
```

PR #1030, opened 2020-11-05
Removed the force unwrap in the FailableIterator extension for the ne…

denis-couras:
…xt () method

Although the method accepts an optional return, the function forces unwrap

### DIFF
--- a/SQLite.xcodeproj/project.pbxproj
+++ b/SQLite.xcodeproj/project.pbxproj
@@ -1257,6 +1257,7 @@
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
 				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks";
+				MARKETING_VERSION = 0.12.3;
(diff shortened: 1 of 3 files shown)


NathanFallet (approved):

jberkel:
As it has been pointed out in the apollographql/apollo-ios#1658, this is a substantial change which has consequences for existing users, so this behaviour change should be documented (and perhaps reverted, if it is not wanted).

groue:
Thanks for noticing, @jberkel. Let's give some time to the Apollo team to understand the consequences (cc @designatednerd), and act accordingly if they think this is necessary.

But basically, yes, this PR has `prepare()` return an untrustable sequence.

- Did you just fetch 10 rows? Well, who knows, maybe there were 100.
- Did you setup a foreign key and a not null constraint so that you are sure all books link to their author? Well, you may load a nil author for some books nevertheless, without any error reporting.
- ...

In my opinion, this change is really concerning, due to the many apps *and other libraries* out there that use SQLite.swift. I'm not sure it is reasonable to expect all of them to audit the consequences and switch to `failableNext()` if necessary. Instead, many app and other libraries now suffer from latent bugs that are just waiting to express.

Before this PR was merged and shipped on 0.13.0, apps and libraries would *crash*. `prepare()` did not hide errors. People has an incentive to switch to `failableNext()`. This was, arguably, a better situation.

PR #1035, opened 2020-12-06
Fix non-standalone lookup

DivineDominion:
This seems to diverge from the upstream/original code, and it makes compiling the project on it's own not work as well. `sqlite3` is only found and imported when e.g. compiling using the CocoaPods settings; otherwise, its `SQLite3`. That's case-sensitive. By removing the module name prefix, we leave figuring this out to the compiler.

Tested this in Timing using CocoaPods on a local clone of the repo.


DivineDominion:
Sorry everyone, picked the wrong PR target 🙇

Issue #897, opened 2019-04-01
How to insert multiple record at once


sahabe1:
It is possible to insert array in once

Issue #168, opened 2015-08-03
INSERT INTO multiple values


krzyzanowskim:
I'd love to insert many values in single query so I build array of Setter

```swift
var keywordsInserts:[Setter] = [Setter]()
keywordsInserts.append(Word <- "A")
keywordsInserts.append(Word <- "B")
keywordsInserts.append(Word <- "C")

Keywords.insert(keywordsInserts) as Int64?
```

and this produced that query:

```sql
INSERT INTO "Keywords" ("Word", "Word", "Word") VALUES ("A","B","C")
```
while expected is 

```sql
INSERT INTO "Keywords" ("Word") VALUES ("A"),("B"),("C")
```

I must doing something wrong. What should I do to get right query?

stephencelis:
This sounds great, but we'd need some kind of way to build a type-safe way of handling this:

``` swift
keywords.insert(Word) { inserts in
    for word in ["A", "B", "C"] {
        inserts.append(word)
    }
}
```

Something like the above. The big issue with most of the type-safe shims in SQLite.swift is the translation layers needed between `Expression<T>`, `Expression<T?>`, `T`, and `T?`. I haven't figured out a scalable way to define a variadic list of expressions and have the compiler to know to use the unwrapped values elsewhere. Because of this, each set of arguments must be handled individually, twice per argument.

```
func insert<A : Value>(Expression<A>, builder: BulkInsertBuilder<A> -> Void)
func insert<A : Value>(Expression<A?>, builder: BulkInsertBuilder<A?> -> Void)
func insert<A : Value, B : Value>(Expression<A>, Expression<B>, builder: BulkInsertBuilder<(A, B)> -> Void)
func insert<A : Value, B : Value>(Expression<A>, Expression<B?>, builder: BulkInsertBuilder<(A, B?)> -> Void)
func insert<A : Value, B : Value>(Expression<A?>, Expression<B>, builder: BulkInsertBuilder<(A?, B)> -> Void)
func insert<A : Value, B : Value>(Expression<A?>, Expression<B?>, builder: BulkInsertBuilder<(A?, B?)> -> Void)
```

It grows exponentially by argument count. If there were a way to define it once and have the type system resolve the wrapping/unwrapping, this would work great. As it stands, I can't figure out a scalable solution.

lukescott:
Perhaps this would work:
```swift
keywords.insert(rows) { row in
    return [Word <- row]
}
```
The function would look like:
```swift
[...]

PR #1048, opened 2021-04-30
implements batch insert rows, insertMany()

geoffmacd:
Feature improvement: ability to insert multiple rows in one statement. 

https://github.com/stephencelis/SQLite.swift/issues/168

This is something I personally need at dropbox for performance as its more efficient to bind to a single statement rather than recreating many similar INSERT Statements. In my use case, using this improves performance for 1 column string insert from ~0.34 seconds for 10,000 rows in a transaction to ~0.08s for 10,000 rows which is inline (to order of magnitude) with other sqlite benchmarks across the internet like https://medium.com/@JasonWyatt/squeezing-performance-from-sqlite-insertions-971aff98eef2. 

Supports all the variations I could see including `Encodable`s and the various syntaxes. There is a slight concern which is that this depends on at least 1 object to find the columns, otherwise, we have to default to some return that makes the API easy to use. eg. it should not return an `Insert?`. Suggestions welcome here. Added tests as well.

Example usage:
- insert many with 1 column: `users.insertMany([mail <- "alice@mac.com"], [email <- "geoff@mac.com"])`


geoffmacd:
tests pass locally, somethings not right with Travis CI though

NathanFallet:
Fixes #897

PR #1053, opened 2021-06-23
Xcode 12.5

jberkel:


jberkel:
@stephencelis the old travis.org service is no longer availble (https://blog.travis-ci.com/2021-05-07-orgshutdown).

I think you need to activate something in order to move the CI to travis-ci.com

Issue #225, opened 2015-09-22
How to alias selected columns in swift 2 ?


huynhphan89:
On branch swift_2.0, alias func was removed in Expression.swift. I dont see any document about the new way to alias columns.

Thanks.

stephencelis:
Aliasing functionality has been limited to tables during the rewrite, where the functionality is required for joins to work. Aliasing column/expressions doesn't seem to help in our expression layer, although I may have missed a use case. Would you please provide your use case?

shnhrrsn:
My use case: I'm selecting all columns in table a, joining table b and selecting just one column from table b that happens to share the name of a column in table a.

If I want to access the that column in table b, it means all of my existing models/queries that reference the column of the same name in table a needs to be updated to support namespacing for just this one scenario.

If I were able to select the column in table b aliased to a different name, all of my existing code would work without modification.  Essentially I'd like the ability to build this query: `SELECT a.*, b.name as otherName FROM a JOIN b ...`

stephencelis:
My experimentation showed that these aliases are actually meaningless outside of the SQLite console. You should be able do do this, though:

``` swift
let a = Table("a")
let b = Table("b")

for row in db.prepare(a.join(b, on: condition)) {
    let aName = row[a[name]]
    let bName = row[b[name]]
}
```

shnhrrsn:
So what I'm actually doing is a little closer to this:

```swift
class Model {
   let name: String

   init(row: Row) {
      self.name = row[name]
   }
}

let a = Table("a")
let b = Table("b")

for row in db.prepare(a.join(b, on: condition)) {
   let model = Model(row)
   // Some how use b[name]
}

```

I initially tried to just update model to do `self.name = row[table[name]]` however queries that weren't joining were failing saying the column didn't exist:

**fatal error: no such column '"a"."name"' in columns: ["\"name\"", ... ]**

If this is _supposed_ to work, than I suppose that solution would be ok, but it would be easier if I could do something along the lines of:

```swift
let a = Table("a")
let b = Table("b")
let otherName = b[name].alias("otherName")

for row in db.prepare(a.select(a[*], otherName).join(b, on: condition)) {
    let model = Model(row)
    let bName = row[otherName]
}
```

This would allow me to isolate this specific bit of code to where I need it, and allow everything else to remain as it is for when namespacing isn't necessary.

mySelfSystem:
[...]

PR #1080, opened 2021-09-02
Fix/column alias

adamwulf:
Fixes https://github.com/stephencelis/SQLite.swift/issues/225 by allowing columns to be aliased. Also allows QueryClauses to be accessed publicly, though only modified internally.


adamwulf:
closed as duplicate of #1081

PR #1081, opened 2021-09-02
Fix/public row

adamwulf:
Allows `Row` objects to be created publicly. This makes it easy to convert statements prepared from raw SQL and run with Connection.run() into the same shape as Connection.prepare(anyTableObject)

### DIFF
--- a/Sources/SQLite/Typed/Expression.swift
+++ b/Sources/SQLite/Typed/Expression.swift
@@ -106,6 +106,9 @@ extension ExpressionType {
         " ".join([self, Expression<Void>(literal: "DESC")])
     }
 
+    public func alias(_ aliasName: String) -> Expressible {
+        return " ".join([self, Expression<Void>(literal: "AS \"\(aliasName)\"")])
+    }
 }
 
 extension ExpressionType where UnderlyingType: Value {
--- a/Sources/SQLite/Typed/Query.swift
+++ b/Sources/SQLite/Typed/Query.swift
@@ -1169,7 +1169,7 @@ public struct Row {
         }
 
         guard let idx = columnNames[column.template] else {
-            let similar = Array(columnNames.keys).filter { $0.hasSuffix(".\(column.template)") }
+            let similar = Array(columnNames.keys).filter { $0.hasSuffix(".\(column.template)") || $0.hasSuffix(" AS \(column.template)") }
 
             switch similar.count {
             case 0:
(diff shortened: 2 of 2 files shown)


jberkel:
Why are there 2 pull requests?

adamwulf:
I could make them one request if you like, i wasn't sure if both would be accepted

jberkel:
Was just confused because it looked like both PRs had the same changes, but yes, having it in one would make reviewing simpler.

adamwulf:
You're completely right - I had messed up the PR and included commits in both. i've closed out the other PR, as this one has everything I've done. Thanks!

jberkel:
Is your use case aliasing aggregate functions as well? The normal API usage would be something like this:
```swift
let name = Expression<String?>("name")
let age = Expression<Int>("age")

let maxAge = age.max
for row in db.prepare(users.select(maxAge).group(name)) {
    print(row[maxAge])
}
```

And you'd like to write
```swift
let maxAge = age.max.alias("maxAge")
for row in db.prepare(users.select(maxAge).group(name)) {
    print(row[maxAge])
}
```
?

adamwulf:
I'm sending query results to a `process(results)` sort of function that expects to have rows with column names A, B, C, etc. In some cases, i can `SELECT A, B, C from foo` (or `foo.select(A, B, C)`). This gives rows that have known column names. in other cases, i'd like to `SELECT A, B, max(C) AS C GROUP BY A, B` and send those query results to the `process()` function too. without the alias, column C doesn't exist since it's named something else by default.

Specifically, the aggregate is happening inside a `VIEW` that's replicating the structure of the underlying table. I'd like to select from either the table or the aggregate view and get the exact same column names back.

With the existing API, the column name for that aggregate method is `max(C)` instead of just `C`. Being able to control the column name of the resulting rows is helpful for consistency.

jberkel:
The problem is that the string alias loses type safety, so it's not possible to do this:
```swift
let maxAge = age.max.alias("maxAge")

for row in db.prepare(users.select(maxAge).group(name)) {
    print(row[maxAge])
}
```

adamwulf:
ah interesting. i'm using it to define a view and i alias to an existing column name, and then use that column object to fetch, so something like:

```
let myView = view.create(mumbleTable.select(colClock.max.alias("existingColumn"), colStuff).group(colStuff))

... then later ...

for row in db.prepare(myView.select(...)) {
    print(row[existingColumn])
}
```

I'll have time this weekend to look deeper into this and see what to do about the type safety side of it.

liyuan116:
let myView = view.create(mumbleTable.select(colClock.max.alias("existingColumn"), colStuff).group(colStuff))

hi @adamwulf  if you delete mumbleTable then create a new table renamed mumbleTable the view myView is work ok?

gzqyl:
let aliasName = columnName, did a trick, finally I found that I do not need alias

PR #1083, opened 2021-09-13
Added supported iOS version to the Package.swift

obrhoff:
Currently the Package.swift lacks a definition of the supported iOS version. I'm not sure if there is a conflict with the #ifdef for the linux version of this package.


jberkel:
superseded by #1094

PR #1051, opened 2021-05-18
Updates for Xcode 12.5

BennettSmith:
* Updates ruby version used by Travis-CI.
* Updates Xcode image to 12.5 for Travis-CI.
* Removes workaround in run-tests.sh for older simulators.
* Updates Carthage toolchain.
* Updates Package.swift tool version.
* Fixes up SPM build for SQLiteObjc module.

PR #1094, opened 2021-11-03
Align platform versions in SPM manifest and Xcode project

calvincestari:
#### This PR:
1. Adds platform versions to the SPM manifest; we discovered this was missing while debugging https://github.com/apollographql/apollo-ios/pull/2015. This addition should allow SPM to correctly resolve deployment targets.
2. Moves the Xcode deployment versions configuration from each target to the project level. There is some personal preference to this change but I do also believe there is value in being able to set all deployment versions in one place (project) vs. the individual targets where it's easy to miss versions not being the same. I also changed the tvOS target to match Cocoapods, which specifies `9.1`.

_I saw that https://github.com/stephencelis/SQLite.swift/pull/1083 already attempted this but that PR seems dead and the fork has also been deleted. If this one is merged you can close #1083._

#### Question:
1. Is there any particular reason why SQLite.swift `0.13.0` bumped the macOS deployment target to `10.15`? It seems at odds with the iOS deployment target of `9.0` which would have greater support for older APIs.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -52,12 +52,6 @@ package.dependencies = [.package(url: "https://github.com/stephencelis/CSQLite.g
 package.targets = [
     .target(
         name: "SQLite",
-        platforms: [
-            .iOS(.v9),
-            .macOS(.v10_15),
-            .watchOS(.v3),
-            .tvOS(.v9)
-        ],
         dependencies: [.product(name: "CSQLite", package: "CSQLite")],
         exclude: ["Extensions/FTS4.swift", "Extensions/FTS5.swift"]
     ),
--- a/Package.swift
+++ b/Package.swift
@@ -3,6 +3,12 @@ import PackageDescription
 
 let package = Package(
     name: "SQLite.swift",
+    platforms: [
+        .iOS(.v9),
+        .macOS(.v10_15),
+        .watchOS(.v3),
+        .tvOS(.v9)
+    ],
     products: [
         .library(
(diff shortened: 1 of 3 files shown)


jberkel:
Thanks for the PR, tests are failing. Regarding MacOS 10.15, I'm not sure, I see it was changed in #1051. Perhaps it can be reverted.

calvincestari:
Thanks for merging that @jberkel!

@NathanFallet it looks like you made the [change](https://github.com/stephencelis/SQLite.swift/commit/449edebfd6c91c2dd376ee61359b4c4b65032032) to the macOS version in the `0.13.0` release. I've built the project using the 'SQLite Mac' with versions 10.15 down back to 10.10 and I don't get any warnings; do you recall any specifics of that particular change?

jberkel:
Seems to work fine on the CI with `10.10`. If nobody objects, we can just set it back before making the next release. The problem is of course testing this on actual `10.10` systems.

NathanFallet:
@calvincestari You're right, I changed the deployment target instead of the base SDK. Sorry for that mistake.

jberkel:
@NathanFallet ok, I reverted it. It's probably worth doing a new release with the current master soon, can you take charge of it?

calvincestari:
Hey @NathanFallet, no need to apologize. Thanks for the change both of you, I'm looking forward to getting our users onto the latest version.

PR #1105, opened 2022-01-17
Native user_version support in Connection

NathanFallet:


NathanFallet:
@jberkel Test fails on Linux because of something related to encodable, but I don't think that it is related to my changes. Do you have any idea?
```
Test Case 'QueryTests.test_update_encodable_with_nested_encodable' started at 2022-01-17 16:31:02.883
/home/runner/work/SQLite.swift/SQLite.swift/Tests/SQLiteTests/QueryTests.swift:396: error: QueryTests.test_update_encodable_with_nested_encodable : XCTAssertEqual failed: ("UPDATE "emails" SET "int" = 1, "string" = '2', "bool" = 1, "float" = 3.0, "double" = 4.0, "date" = '1970-01-01T00:00:00.000', "sub" = '{"date":-978307200,"int":1,"float":3,"string":"2","double":4,"bool":true}'") is not equal to ("UPDATE "emails" SET "int" = 1, "string" = '2', "bool" = 1, "float" = 3.0, "double" = 4.0, "date" = '1970-01-01T00:00:00.000', "sub" = '{"date":-978307200,"double":4,"int":1,"string":"2","bool":true,"float":3}'") - 
Test Case 'QueryTests.test_update_encodable_with_nested_encodable' failed (0.001 seconds)
```

jberkel:
@NathanFallet the test is brittle because it compares the JSON as a string. On Linux the JSON serialization is different (order of fields). You could make the JSON serialization more deterministic with `OutputFormatting.sortedKeys` (but that's only iOS 11+,  macOS 10.13+) or improve the test so that it does not do a by string comparison.

NathanFallet:
@jberkel Right, I'm gonna inspect that. I need to understand why it changed since the last commit (because test was passing for the last commit on master)

Sources/SQLite/Core/Connection.swift
```diff
@@ -152,6 +152,23 @@ public final class Connection {
         Int(sqlite3_total_changes(handle))
     }
 
+    /// Gets the user version of the database.
+    /// See SQLite [PRAGMA user_version](https://sqlite.org/pragma.html#pragma_user_version)
+    /// - Returns: the user version of the database
+    public func getUserVersion() throws -> Int32? {
```
jberkel, line 158:
```
var userVersion: Int32? {
   get { }
   set { }
}
```

NathanFallet: I tried that first, but I ended with problems with `try`. I can eventually ignore them with optionals

NathanFallet:
> @NathanFallet the test is brittle because it compares the JSON as a string. On Linux the JSON serialization is different (order of fields). You could make the JSON serialization more deterministic with `OutputFormatting.sortedKeys` (but that's only iOS 11+, macOS 10.13+) or improve the test so that it does not do a by string comparison.

After inspection, it means that `JSONEncoder().encode(value)` gives different data each time, even with the same object passed as `value`?!

Sources/SQLite/Core/Connection.swift
jberkel: Yes, I think it's ok to ignore the error here

jberkel:
> After inspection, it means that `JSONEncoder().encode(value)` gives different data each time, even with the same object passed as `value`?!

Possibly, there are no guarantees by the method. I think we should do both: make the JSON output consistent (with `sortedKeys`) if possible, and avoid doing a string comparison in the test.

Sources/SQLite/Core/Connection.swift
```diff
@@ -152,21 +152,20 @@ public final class Connection {
...
-    /// Gets the user version of the database.
+    /// The user version of the database.
     /// See SQLite [PRAGMA user_version](https://sqlite.org/pragma.html#pragma_user_version)
-    /// - Returns: the user version of the database
-    public func getUserVersion() throws -> Int32? {
-        guard let userVersion = try scalar("PRAGMA user_version") as? Int64 else {
-            return nil
+    public var userVersion: Int32? {
+        get {
+            guard let userVersion = try? scalar("PRAGMA user_version") as? Int64 else {
```
jberkel, line 159:
```swift
(try? scalar("PRAGMA user_version") as? Int64).map(Int32.init)
```

Tests/SQLiteTests/TestHelpers.swift
```diff
@@ -125,4 +125,15 @@ class TestCodable: Codable {
         self.optional = optional
         self.sub = sub
     }
+    
+    static func == (lhs: TestCodable, rhs: TestCodable) -> Bool {
```
jberkel, line 129: the compiler can generate this for you

Sources/SQLite/Core/Connection.swift
```diff
@@ -152,6 +152,19 @@ public final class Connection {
...
+    /// The user version of the database.
+    /// See SQLite [PRAGMA user_version](https://sqlite.org/pragma.html#pragma_user_version)
+    public var userVersion: Int32? {
+        get {
+            (try? scalar("PRAGMA user_version") as? Int64).map(Int32.init)
+        }
+        set {
+            if let userVersion = newValue {
+                _ = try? run("PRAGMA user_version = \(userVersion)")
+            }
```
jberkel, line 164: what should happen in the nil case? It's also reasonable to expect that it resets to the default value

Tests/SQLiteTests/TestHelpers.swift
NathanFallet: Works only for structs, not classes. And this class cannot be a struct because it contains a nested object with same class.

Sources/SQLite/Core/Connection.swift
NathanFallet: I'm going to add an else clause to reset it to 0 then

PR #1112, opened 2022-02-18
UUID Fix

benrlewis7450G:
Inserting a UUID as Expression<UUID> resulted in escaped quotes on either side of the uuidString and thus actual quotes stored inside the DB.  When retrieving a row, the result of UUID(uudiString:) returns null when a uuidString with quotes on either side is given.
This appears to be an issue with JSONEncoder.
Fix pre-empts JSONEncoder and inserts a proper uuidString without escaped quotes on either side.

### DIFF
--- a/Sources/SQLite/Typed/Coding.swift
+++ b/Sources/SQLite/Typed/Coding.swift
@@ -207,6 +207,8 @@ private class SQLiteEncoder: Encoder {
                 encoder.setters.append(Expression(key.stringValue) <- data)
             } else if let date = value as? Date {
                 encoder.setters.append(Expression(key.stringValue) <- date.datatypeValue)
+            }else if let uuid = value as? UUID {
+                encoder.setters.append(Expression(key.stringValue) <- uuid.uuidString)
             } else {
                 let encoded = try JSONEncoder().encode(value)
                 let string = String(data: encoded, encoding: .utf8)


jberkel:
Thanks for the PR. Looks like there are still some linting errors.

jberkel:
Install `swiftlint` (`$ brew install swiftlint`) or see https://github.com/realm/SwiftLint, then do 

```$ make lint```

File: Sources/SQLite/Typed/Coding.swift [hunk 1]
```diff
@@ -207,6 +207,8 @@ private class SQLiteEncoder: Encoder {
                 encoder.setters.append(Expression(key.stringValue) <- data)
             } else if let date = value as? Date {
                 encoder.setters.append(Expression(key.stringValue) <- date.datatypeValue)
+            } else if let uuid = value as? UUID {
+                encoder.setters.append(Expression(key.stringValue) <- uuid.uuidString)
```
jberkel, line 211: `<- uuid.datatypeValue`

Tests/SQLiteTests/FoundationTests.swift
```diff
@@ -25,4 +25,30 @@ class FoundationTests: XCTestCase {
...
+        try! db.run(iQuery)
+
+        let fQuery = table.filter(uuid == testUUID)
+        if let result = try! db.pluck(fQuery) {
+            let testValueReturned = Test(uuid: result[uuid], string: result[string])
+            XCTAssertEqual(testUUID, testValueReturned.uuid)
+        } else {
+            XCTFail("Search for uuid failed")
+        }
+    }
```
jberkel, line 53: the tests should be in `QueryTests.swift`. Look for `test_insert_encodable`

PR #1119, opened 2022-03-17
Add prepareRowIterator method to an extension of Statement.

jjamminjim:
Added unit-test.
Updated Documentation.


jberkel:
What's the use case of this? It's duplicating the `prepareRowIterator` functionality in `Connection`.

Issue #1133, opened 2022-05-09
Should be able to access columns by index, not just by key.


justinmeiners:
## Overview

Suppose a query result returns a large number of rows:

    let rows = try db.prepare(...)

To access the data, I have to loop through each row, and on each row lookup the desired columns by key. This requires a call to `Row.get` which does a lot of work (quoting the column name, looking it up in a dictionary, checking for ambiguity, etc).

     for row in rows {
         let x = row.get(Expression<double>("value"))
     }

This is a lot of needless work, when each row contains the exact same information.
Alternatively, I should be able to look up the index of the column first, and then simply access it from each row.
Here is a mockup of what this might look like:

    let valueIndex = rows.first!.findIndex("value")

    for row in rows {
          let x: Double = row.get(valueIndex)
    }

## Motivation

This is killing my performance when decoding a large number of rows (20,000 with 10 columns each).

This suggestion also better matches how the SQLite C api works, and most SQLite libraries (such as Python's standard library).

## Possible implementation

I think we can get there with a few small changes:

1. Refactor `Row.get` into two functions: `Row.get` and `Row.getIndex`.
    `getIndex` will have almost all of the code currently there.
    `get` can then immediately call `getIndex`.

3. Make `Row.values` public.

**Alternative**

It's possible that the SQLiteEncoder and SQLiteDecoder could modified to decode arrays of codable objects. This could then efficiently do something like described above.

justinmeiners:
This can be done using a raw query and returning a `Statement`.

PR #1134, opened 2022-05-09
Initial implementation of indexed column access

justinmeiners:
See #1133. This is just an initial implementation for feedback.

### DIFF
--- a/Sources/SQLite/Typed/Query.swift
+++ b/Sources/SQLite/Typed/Query.swift
@@ -1162,13 +1162,10 @@ public struct Row {
         }
     }
 
-    public func get<V: Value>(_ column: Expression<V?>) throws -> V? {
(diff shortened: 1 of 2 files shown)


justinmeiners:
@jberkel Due to time constraints, I may need to contribute this in an internal fork to support my work. I think we would both benefit if that work goes back into this project, rather than diverging. Can you provide any early guidance here to make sure I meet your expectations and stay on track?

Does the general idea seem sound to you?

jberkel:
Is this part of a performance optimization? In general I don't see issues but a user of the library might be confused which method to use when accessing row data. Also, there's a bit of type loss when going from `Expression<V>` to `Int`.

justinmeiners:
@jberkel I figured out how to access what I needed through the lower level statement object.

Issue #1126, opened 2022-04-22
UUID fields are not decoded properly (while encoded good)


Alkenso:
Version: 0.13.3
Installed via SPM

SQLite/Typed/Coding.swift
SQLite.swift has support of UUID type.
- it is encoded properly at line 210
- decoding counterpart is missing at line 390

eonil:
+1

Same thing happened to me.

PR #1137, opened 2022-06-10
add decoding for UUID

atultw:
I stumbled upon issue #1126 while trying to decode a UUID using codable. As described there, encoding worked fine but decoding treated the field like json.

Added an `else` to cover the UUID case, matching the `encode` method functionality.

### DIFF
--- a/Sources/SQLite/Typed/Coding.swift
+++ b/Sources/SQLite/Typed/Coding.swift
@@ -387,7 +387,11 @@ private class SQLiteDecoder: Decoder {
             } else if type == Date.self {
                 let date = try row.get(Expression<Date>(key.stringValue))
                 return date as! T
+            } else if type == UUID.self {
+                let uuid = try row.get(Expression<UUID>(key.stringValue))
+                return uuid as! T
             }
+            
             // swiftlint:enable force_cast
             guard let JSONString = try row.get(Expression<String?>(key.stringValue)) else {
                 throw DecodingError.typeMismatch(type, DecodingError.Context(codingPath: codingPath,


atultw:
lint should pass now

jberkel:
Thanks for your contribution. Can you add some tests?

atultw:
Added a UUID field to `TestCodable`. I figure that's the best way since it already has fields for date and other types. 

I copy pasted apple's example UUID from foundation docs for the tests. May want to make it a constant instead. Let me know what you think!

Sources/SQLite/Typed/Coding.swift
```diff
@@ -387,7 +387,11 @@ private class SQLiteDecoder: Decoder {
             } else if type == Date.self {
                 let date = try row.get(Expression<Date>(key.stringValue))
                 return date as! T
+            } else if type == UUID.self {
+                let uuid = try row.get(Expression<UUID>(key.stringValue))
+                return uuid as! T
             }
+
```
jberkel, line 394: As the number of branches get bigger, it might be better to use a switch statement here:
```swift
switch type {
case is Data.Type:
...
```

The same actually for encode:
```swift
func encode<T>(_ value: T, forKey key: Key) throws where T: Swift.Encodable {
    switch value {
    case let data as Data:
    ...
```

Tests/SQLiteTests/QueryIntegrationTests.swift
```diff
@@ -75,15 +75,15 @@ class QueryIntegrationTests: SQLiteTestCase {
...
             builder.column(Expression<Double>("double"))
             builder.column(Expression<Date>("date"))
+            builder.column(Expression<UUID>("uuid"))
             builder.column(Expression<String?>("optional"))
             builder.column(Expression<Data>("sub"))
         })
 
         let value1 = TestCodable(int: 1, string: "2", bool: true, float: 3, double: 4,
-                                 date: Date(timeIntervalSince1970: 0), optional: nil, sub: nil)
+                                 date: Date(timeIntervalSince1970: 0), uuid: UUID(uuidString: "E621E1F8-C36C-495A-93FC-0C247A3E6E5F")!, optional: nil, sub: nil)
```
jberkel, line 84: should be defined once in a  variable

atultw: added to TestHelpers

atultw:
Thanks Jan. Made the changes, tests passing. I put the uuid constant in `TestHelpers`, can move it elsewhere if needed

atultw:
Tried splitting the lines that were too long into two lines, but that made the test fail. So I disabled the rule at those spots.

PR #1151, opened 2022-08-30
FIX index.md

Gobans:
'pathForResource(_:ofType:)' has been renamed to 'path(forResource:ofType:)'


NathanFallet (approved):
That's right, thanks for your contribution

PR #1164, opened 2022-10-18
Run  CI on macOS 12 (do not merge)

jberkel:

PR #1165, opened 2022-10-18
Remove redundant assignment (do not merge)

jberkel:


NathanFallet:
So is #1164 included in this PR?

PR #1167, opened 2022-10-23
Reactivate "Performance improvements for Blob"

jberkel:
#416

### DIFF
--- a/Sources/SQLite/Core/Blob.swift
+++ b/Sources/SQLite/Core/Blob.swift
@@ -40,17 +40,7 @@ public final class Blob {
             self.init(data: NSData())
             return
         }
-        let buffer = UnsafeMutablePointer<UInt8>.allocate(capacity: bytes.count)
-        for idx in 0..<bytes.count {
-            buffer.advanced(by: idx).pointee = bytes[idx]
-        }
-
-        let data = NSData(
-            bytesNoCopy: UnsafeMutableRawPointer(buffer),
-            length: bytes.count,
-            freeWhenDone: true
-        )
-        self.init(data: data)
+        self.init(data: NSData(bytes: bytes, length: bytes.count))
     }
 
     public convenience init(bytes: UnsafeRawPointer, length: Int) {
--- a/Tests/SQLiteTests/Core/BlobTests.swift
+++ b/Tests/SQLiteTests/Core/BlobTests.swift
@@ -13,9 +13,19 @@ class BlobTests: XCTestCase {
         XCTAssertEqual(blob.toHex(), "")
     }
 
+    func test_description() {
+        let blob = Blob(bytes: [0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 150, 250, 255])
+        XCTAssertEqual(blob.description, "x'000a141e28323c46505a6496faff'")
+    }
+
+    func test_description_empty() {
+        let blob = Blob(bytes: [])
+        XCTAssertEqual(blob.description, "x''")
+    }
+
     func test_init_array() {
-        let blob = Blob(bytes: [42, 42, 42])
-        XCTAssertEqual([UInt8](blob.data), [42, 42, 42])
+        let blob = Blob(bytes: [42, 43, 44])
+        XCTAssertEqual([UInt8](blob.data), [42, 43, 44])
     }
 
     func test_init_unsafeRawPointer() {


Sources/SQLite/Core/Blob.swift
```diff
@@ -21,26 +21,50 @@
...
-public struct Blob {
+public final class Blob {
 
-    public let bytes: [UInt8]
+    public let data: NSData
 
-    public init(bytes: [UInt8]) {
-        self.bytes = bytes
+    public var bytes: UnsafePointer<UInt8> {
+        data.bytes.assumingMemoryBound(to: UInt8.self)
```
stephencelis, line 31: This is unfortunately undefined behavior and could lead to crashes. We should revert this.

Sources/SQLite/Core/Blob.swift
```diff
@@ -21,26 +21,50 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 //
+import Foundation
 
-public struct Blob {
+public final class Blob {
 
-    public let bytes: [UInt8]
+    public let data: NSData
```
stephencelis, line 28: I think these days `NSData` is going to make things slower in Swift. If we want to make fundamental changes like this they should be benchmarked. I have some other repos with examples that use `swift-benchmark` if that would help!

stephencelis (commented):
@jberkel Just caught wind of this change and I think it should be reverted quickly. It introduces a serious security issue for folks accessing the bytes interface.

stephencelis:
@jberkel Sorry for the fly-by! And much appreciation for your continual maintenance of this :smile: Just don't want folks to get bit by it in production.

stephencelis:
@jberkel I just pulled the 0.14.0 tag to prevent folks from taking this change for now, but please cut 0.14.1 when you have a chance to revert.

stephencelis:
@jberkel Again, sorry for the fly-by out of nowhere! Please let me take some time to try to flesh out my concerns. I'll start with the one that I think is the most pressing:

1. Adding an API that returns an unsafe, dangling pointer:

    ```swift
    public var bytes: UnsafePointer<UInt8> {
      data.bytes.assumingMemoryBound(to: UInt8.self)
    }
    ```

    Dangling pointers can easily outlive the lifetime of the underlying memory and be deallocated before use. Swift provides scoped methods like `withUnsafePointer` that guarantee the lifetime of the pointer for the duration of the scope for a reason, but we're not doing the same here. And while `NSData` may copy the bytes from the original SQL result, I'm more worried that this `NSData` is the owner of the underlying memory and could deallocate the pointer before a SQLite.swift user interacts with it. This could lead to hard-to-diagnose crashes and other memory issues, and easily become an attack vector in Swift apps using SQLite.swift.

    Also note that `Data`, the Swift-friendly wrapper around `NSData`, does not provide an interface to this dangling pointer because it is not safe.

    I think Swift libraries should avoid exposing `Unsafe*` APIs like this in general, and when they do, the reason should be clearly documented and motivated. This is even a prerequisite of libraries Apple accepts into [SSWG incubation](https://github.com/swift-server/sswg/blob/master/process/incubation.md#minimal-requirements). I hope you agree that we should probably follow suit.

2. Updating the API to use a new data structure for performance improvements without validating the improvements with benchmarks. While it may be true that `NSData` and `Data` were anecdotally more performant when the original PR was opened six years ago, I'm pretty sure this is no longer the case today. Swift's performance has improved a lot since then, and while there may be opportunities to improve the library performance, I do think that it should be measured in actual benchmarks. And in the case of this change, I worry that it may have a measurably negative impact on library users.

3. I don't think we should assume the underlying memory of `NSData.bytes` without very strong guarantees, and in general I think `assumingMemoryBound(to:)` should be avoided if there are safer alternatives. While the library may have other uses of `assumingMemoryBound(to:)`, those should probably be audited to something safer, or the use should be clearly motivated/documented.

Hope these concerns are clear and make sense! I appreciate wanting to continually improve the library for its users, but I also want to make sure we don't ship something that could cause unintentional problems for them.

stephencelis:
> It's only a dangling pointer once the `Blob` / `NSData` is deallocated

Sure, but APIs that return dangling pointers make it very easy for folks to make this mistake by holding onto the pointer and letting the data be deallocated.

> This is probably the weakest argument, as stated above I don't see how this behavior can be undefined according to the Swift memory model.

I'd like to be a little more cautious here as I don't see how we can guarantee it. We know that using unsafe interfaces can cause problems at runtime when used incorrectly. Might be worth asking the compiler engineers on the Swift forums if it is safe to use this API here and, if so, document the findings?

PR #1179, opened 2022-12-23
Fix incorrect behavior when preparing SELECT * preceded by a WITH

marmphco:
In  #1139 I introduced support for the `WITH` clause. My implementation contains a bug: preparing a query containing a `SELECT *` preceded by a `WITH` causes an error to be thrown.

Consider the following statement:

```sql
WITH temp AS ( SELECT id, email from users) SELECT * from temp
```

An error is thrown when preparing this statement because the glob expansion procedure tries to determine the column names for the result by looking up the column names for the query `SELECT * from temp`. This does not work because `temp` is a temporary view defined in the `WITH` clause.

To fix this, I modified the glob expansion procedure to include the `WITH` clause in the query used to look up the result column names.

### DIFF
--- a/Sources/SQLite/Typed/Query.swift
+++ b/Sources/SQLite/Typed/Query.swift
@@ -1036,10 +1036,29 @@ extension Connection {
             let column = names.removeLast()
             let namespace = names.joined(separator: ".")
 
+            // Return a copy of the input "with" clause stripping all subclauses besides "select", "join", and "with".
+            func strip(_ with: WithClauses) -> WithClauses {
+                var stripped = WithClauses()
+                stripped.recursive = with.recursive
+                for subclause in with.clauses {
+                    let query = subclause.query
+                    var strippedQuery = type(of: query).init(query.clauses.from.name, database: query.clauses.from.database)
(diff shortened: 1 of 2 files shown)


Sources/SQLite/Typed/Query.swift
```diff
@@ -1036,10 +1036,29 @@ extension Connection {
...
+                    stripped.clauses.append(strippedSubclause)
+                }
+                return stripped
+            }
+
             func expandGlob(_ namespace: Bool) -> (QueryType) throws -> Void {
                 { (queryType: QueryType) throws -> Void in
                     var query = type(of: queryType).init(queryType.clauses.from.name, database: queryType.clauses.from.database)
                     query.clauses.select = queryType.clauses.select
+                    query.clauses.with = strip(queryType.clauses.with)
```
marmphco, line 1061: The `strip` function feels a little excessive. Doing this instead also works:
```swift
query.clauses.with = queryType.clauses.with
```

but the query will include things that aren't necessary for determining the result column names, like filters, sorting, and grouping.

NathanFallet (approved):

PR #1202, opened 2023-05-01
Bump iOS minimum deployment target to 11.0

danielob:
I've noticed that when opening the project with Xcode 14.3, we get a warning that the target is not in the range of supported versions. Additionally, when trying to build the project using Carthage, we get a build error due to the deployment target being set to 8.0.

To address this issue, I've made changes to update the deployment target version to 11.0, which is within the range of supported deployment target versions for both Xcode and Carthage. This will ensure that the project can be built and run smoothly with the latest versions of Xcode and Carthage.

I've tested the changes locally and confirmed that they resolve the warning and error messages. Please review my changes and let me know if you have any feedback or concerns.


NathanFallet:
@danielob @jberkel Tests are not passing, I don't understand why...

PR #1217, opened 2023-06-07
SchemaReader: return the correct column definition for a composite primary key

stefansaasen:
The `PRAGMA` `table_info` that is used to return the column definitions, returns one row for each defined column. The `pk` column contains:

> ... either zero for columns that are not part of the primary key, or the 1-based index of the column within the primary key).

See https://www.sqlite.org/pragma.html#pragma_table_info

Checking whether the `pk` column equals 1 only detects a single primary key and ignores other columns that are part of a composite primary key.


NathanFallet (approved):
LGTM, thanks!

PR #1248, opened 2024-03-05
Fix #1247 support nil case when decoding optionals

### LINKED ISSUE #1247: bug in optional support for decoding
If a table includes an optional custom struct, the decode method throws an error when the optional value is nil

I created a unit test demonstrating this here:
https://github.com/stephencelis/SQLite.swift/compare/master...Chris-Stockbridge:SQLite.swift:master
d5b2adc48a2a458b6e366d39a1ffa060381c3912

This unit test passes with the previous release, but breaks starting with 0.15.0

Chris-Stockbridge:
#1247 describes a bug where decoding a nil value in an optional throws an error when nil is a valid response.

This same bug occurs with arrays or structs encoded to either optional string or data types.

## Testing
I added a unit test "test_insert_custom_encodable_type" which demonstrates the issue. The current master branch fails when decoding a nil value.

### DIFF
--- a/Sources/SQLite/Typed/Coding.swift
+++ b/Sources/SQLite/Typed/Coding.swift
@@ -504,9 +504,8 @@ private class SQLiteDecoder: Decoder {
             case is UUID.Type:
                 return try? row.get(Expression<UUID>(key.stringValue)) as? T
             default:
-                guard let JSONString = try? row.get(Expression<String?>(key.stringValue)) else {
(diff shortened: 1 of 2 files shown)


nathanfallet (approved):
You're right! `decodeIfPresent` was added not a long ago, and that's its goal to return `nil` instead of throwing. I missed that point in the review.

PR #1264, opened 2024-04-18
Update project file to reflect correct marketing versions

SagarSDagdu:
## What does this PR do

For recent releases, the `Info.plist` shipped inside the `xcframeworks` does not match the release version. Fixing that to reflect the latest marketing version.

@nathanfallet should we think of a better approach of updating this every time a release is created? (like a script that does that)


SagarSDagdu:
Closing this as already added these commits in other linked PR

PR #1273, opened 2024-07-22
Fix force try in the `FailableIterator.next()` and `Connection.prepare()` methods

syoung-smallwisdom:
The force try is crashing and ideally, the prepare() query should throw the error rather than crashing the app. Thanks! 

### DIFF
--- a/Sources/SQLite/Core/Statement.swift
+++ b/Sources/SQLite/Core/Statement.swift
@@ -214,8 +214,8 @@ public protocol FailableIterator: IteratorProtocol {
 
 extension FailableIterator {
     public func next() -> Element? {
-        // swiftlint:disable:next force_try
-        try! failableNext()
+        // Do not force try 
+        try? failableNext()
     }
 }
 
--- a/Sources/SQLite/Typed/Query.swift
+++ b/Sources/SQLite/Typed/Query.swift
@@ -1009,10 +1009,9 @@ extension Connection {
         let statement = try prepare(expression.template, expression.bindings)
 
         let columnNames = try columnNamesForQuery(query)
+        let rows = try statement.failableNext().map { Row(columnNames, $0) }
 
-        return AnySequence {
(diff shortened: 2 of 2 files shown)


jokerYellow:
anyone please merge this request?

dyikai:
Why don't you just use `prepareRowIterator` instead of `prepare`?

syoung-smallwisdom:
> Why don't you just use `prepareRowIterator` instead of `prepare`?

I don't understand the question. The `prepare()` method is a publicly exposed method that is included in the [README](https://github.com/stephencelis/SQLite.swift/blob/c7b78f2f9a0376bc0145f3ecbd0df05eab1688f7/README.md?plain=1#L64) as an example usage of the library.

dyikai:
I notice this line was reverted from `try?` in the past, this may be the reason #1075 and the previous fix pr is #1030

syoung-smallwisdom:
> I notice this line was reverted from `try?` in the past, this may be the reason #1075 and the previous fix pr is #1030

If you read through the comments for _why_:
#1030 - Don't crash _and_ documentation should be updated.
#1075 - Revert the change b/c it causes an error in `prepare()`.

Note that the documentation was never updated to tell users that "If you use this method that we included in the library and use in the example README, then it can cause your production app crash". Seems to me that there are two solutions to this:

1. Mark the method as unavailable *and* don't implement the `Iterator` protocol.
*OR*
2. Fix it by making the `next()` method adhere to the protocol *and* fixing `prepare()` (which can throw errors) to use `failableNext()` instead.

I chose Option 2. It's up to those who still use this fork of the repo to decide if they want to accept the fix - I would certainly suggest that if the community would like to keep people from forking off and maintaining their own forks, then "just use a different method that is not in the documented README" is not a solution. 

FWIW, anyone who wants this fix is welcome to use my client's fork: https://github.com/BiAffectBridge/SQLite.swift

xiushaomin:
any update?

arthursalgado:
Any update?

Issue #1317, opened 2025-05-27
cross compilation for linux on macOS fails


ha100:
compilation fails with

```
⋮
.build/checkouts/SQLite.swift/Sources/SQLite/Core/Backup.swift:32:8: error: no such module 'CSQLite'
 30 | import SQLCipher
 31 | #elseif os(Linux)
 32 | import CSQLite
    |        `- error: no such module 'CSQLite'
 33 | #else
 34 | import SQLite3
⋮
```

when invoked with command `swift build --swift-sdk x86_64-swift-linux-musl` due to `CSQLite` dependency being hidden behind preprocessor macro

```
#if os(Linux)
⋮
#endif
```

perhaps it could be protected by ENVIRONMENT variable

```
let swiftTarget = ProcessInfo.processInfo.environment["SWIFT_BUILD_TARGET"]
⋮

if swiftTarget == "linux" {
⋮
}
```

PR #1318, opened 2025-06-08
fix cross compilation for linux on macOS fails #1317

ha100:
allows SQLite.swift to be used for cross compilation with `swift build --swift-sdk x86_64-swift-linux-musl` command


nathanfallet:
Thank you!

Issue #1322, opened 2025-07-09
Add overload of createIndex that takes an array instead of variadic parameters


bbrk24:
Currently, the `Table.createIndex` method only has one overload:

```swift
public func createIndex(_ columns: Expressible..., unique: Bool = false, ifNotExists: Bool = false) -> String
```

It would be nice to have a second overload like
```swift
public func createIndex(_ columns: [Expressible], unique: Bool = false, ifNotExists: Bool = false) -> String
```

so that it becomes possible to wrap it in a helper function that can take an arbitrary number of columns in its parameters. Currently, there's no way to forward variadic arguments.

PR #1326, opened 2025-07-17
Create index support array as parameters to solve #1322

hawk0620:

overload `public func createIndex(_ columns: [Expressible], unique: Bool = false, ifNotExists: Bool = false) -> String` and `public func dropIndex(_ columns: [Expressible], ifExists: Bool = false) -> String` to solve [#1322](https://github.com/stephencelis/SQLite.swift/issues/1322)


nathanfallet (approved):
LGTM 👍
Thank you!

PR #1327, opened 2025-07-19
Added support for casting when Codable.

miku1958:
Although the data stored in SQLite has specified types, these are more for guidance. If an int value is stored as a string, it should be readable as an int. This is the behavior supported by SQLite.

# Changes
- Use LazySequence to delay the creation of Cursor (statement.handle) and the access of Cursor.getValue to match the variable type of Codable.
- Change subscribe of Cursor to a regular method with throws to bypass the compiler error Ambiguous use of 'xxxxx' and simplify the indentation of get throws.

### DIFF
--- a/Sources/SQLite/Core/Statement.swift
+++ b/Sources/SQLite/Core/Statement.swift
@@ -166,7 +166,7 @@ public final class Statement {
 
         reset(clearBindings: false)
         _ = try step()
-        return row[0]
+		return try row.getValue(0)
     }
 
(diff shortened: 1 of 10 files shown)


nathanfallet:
Thanks for your PR! Tell me when it's ready so I can review it and merge it (I saw you're still pushing some code)

miku1958:
Interesting, SQLite.xcodeproj is not using the SPM

nathanfallet:
@miku1958 that's because of Cocoapods I think

Issue #1071, opened 2021-08-25
Custom functions broken on Linux


jberkel:
Caused by this bug in Swift:

https://bugs.swift.org/browse/SR-4429

PR #1330, opened 2025-08-13
Attempt to fix tests build for android

NikolayJuly:


nathanfallet:
I'm launching them. A maintainer needs to approve them for for new contributors to run.

NikolayJuly:
Looks like same bug,  as for linux - https://github.com/stephencelis/SQLite.swift/issues/1071 . Which make sense... Pushing fixes.

NikolayJuly:
@nathanfallet I'm not familiar with setup of test for iOS - but my changes can't break it.
May be you can point to  a right direction?
Error message:
```
xcodebuild: error: Unable to find a device matching the provided destination specifier:
		{ platform:iOS Simulator, OS:18.0, name:iPhone 16 }
```

nathanfallet:
@NikolayJuly Try opening Xcode to launch manually an iPhone 16 on iOS 18.0 simulator, so it's created once and then it will find it. If you updated your Mac to a more recent version, you might have to manually download it from the "Windows & Simulator" menu.

marcprux:
> Looks like iOS CI start failing here - #1326 But I think, it has nothing to do with that PR, honestly. It just looks like some step misconfigured on CI.

The iOS CI failure is due to https://github.com/actions/runner-images/issues/12758#issuecomment-3206748945

nathanfallet:
I get it. So we should update the iOS tests to use the latest available iOS instead of a fixed version maybe?

marcprux:
> I get it. So we should update the iOS tests to use the latest available iOS instead of a fixed version maybe?

As I understand it, the options are to:

1. Override the version of Xcode in use (e.g., to 16.4 or 26)
1. Downgrade the runner to macos-14 (which didn't suffer this degradation), or 
1. Wait until August 25th when the default Xcode installed on the macos-15 instances will be bumped from 16.0 to 16.4 (https://github.com/actions/runner-images/issues/12751).

NikolayJuly:
Option 3 looks okay for me, especially since I'm now trying to push my package to production on Android. I might have more info/ideas in 5 days.
But honestly, it is disappointing to see that GitHub forces such compromises and doesn’t work out of the box.

@marcprux thank you for digging up this info!

NikolayJuly:
@nathanfallet I've seen initial failure, I also notice that device list first time didn't contain any actual sim. And on second run there was big list of simulators.... But looks like latest changes actually trigger test run on sim. 

May be worth consider later for separate PR smth like take list from `$ xcrun simctl list runtimes` and jsut parse latest iOS sim... Cos idea of harcoding iOS version and device - looks strange for me.

nathanfallet:
@NikolayJuly yes it looks like it works with the latest one.

We might have to either keep updating it manually or have it removed/automatically select it so we don't have it broken.

At least I'll merge this one if it works because the initial goal was to fix android build. But if you feel brave enough to try this we can open a new PR for it just after merging this one.

PR #1332, opened 2025-08-27
Added Sendability conformance

samdogg7:
Added Sendability conformance, where I could without modifying underlying code.

### DIFF
--- a/Sources/SQLite/Core/Blob.swift
+++ b/Sources/SQLite/Core/Blob.swift
@@ -22,7 +22,7 @@
 // THE SOFTWARE.
 //
 
-public struct Blob {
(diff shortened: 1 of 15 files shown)


samdogg7:
@nathanfallet I am wrapping SQLite.swift in my own module that I'd like to migrate over to Strict Concurrency. In my case, I particularly would like to mark `Expression` as `Sendable`. I added in a variety of `Sendable` conformances across the repository that didn't require changes to the underlying code. Larger coverage would be beneficial, but I think can be done iteratively.

samdogg7:
@nathanfallet ah my bad, I've pulled in latest changes

PR #1356, opened 2026-03-19
TR-4764: Forward-port to upstream 0.16.0

Jeehut:
I went with a **forward-port** approach — starting from upstream 0.16.0 and layering our fork-specific patches back on top, rather than trying to rebase ~20 custom commits through upstream's trait system refactoring. Much cleaner this way.

**What we pick up from upstream:** Sendable conformances across the type system (Binding, Blob, Expression, QueryClauses, Operator, etc. — good for Swift 6 prep), PrivacyInfo.xcprivacy as an SPM resource, a richer Schema module, and assorted fixes.

**What should stay the same for Timing:** Epoch Double date storage, the SafeValue/RiskyValue type split, Result-wrapped row iteration, the SQLiteSwift module rename, non-optional Bool expressions, FTS4 tokenizer quoting, and the TR-2083 rollback suppression. Claude went through all 37 source files organized into patch clusters, and Codex reviewed the result across two review rounds — as far as I can tell, none of the Timing-specific behavior should have changed.

A couple of extras fell out along the way — the test target had 33 files still importing the old `SQLite` module name, so those got fixed and all **480 tests pass** now. There was also an old `sqlite3_trace` fallback that doesn't exist in our CSQLite build (removed), and one standalone import path that Codex caught pointing at CSQLite instead of sqlite3 (fixed). I also added an `.editorconfig` and normalized stray tabs to spaces to match upstream convention.

**On whether we should reduce the fork further:** I also looked at whether this was the right moment to collapse more of the fork back toward upstream. I don't think that belongs in this PR — the remaining differences cover on-disk date format, app-facing protocols, iteration safety, and Bool-expression shape, and changing any of those deserves its own scope and testing.

That said, the remaining divergences aren't all equal in terms of maintenance cost:

- **Permanent by design:** epoch Date storage, SQLiteSwift module rename, and probably SafeValue/RiskyValue — these are tied to stored data and deeply used app code
- **Low-risk cleanup candidates:** Bool non-optionality and FTS4 quoting could likely be dropped or upstreamed with minimal impact; CSQLite base alignment is possible packaging cleanup
- **Needs a deliberate follow-up decision:** the FailableIterator Result-wrapping is the biggest source of friction during upgrades, but it provides real crash safety on corrupt rows — worth discussing separately whether we can get that safety at a different layer

Happy to create a follow-up task for any of these if you want to pursue them.


Jeehut:
Opened against wrong repo by mistake, meant for the Timing fork.

PR #1361, opened 2026-05-31
Add explicit deterministic close support for SQLite connections

RuiNelson:
This PR adds an explicit `Connection.close()` API so callers can deterministically close a SQLite database connection before releasing or deleting the underlying database file.

**Motivation**

Some applications need to manage short-lived SQLite database files explicitly. In those cases, relying only on `Connection.deinit` can make file lifetime difficult to coordinate with external cleanup.

This is especially important when a connection may still have active SQLite resources, such as prepared statements. SQLite’s `sqlite3_close()` returns `SQLITE_BUSY` in that case and keeps the connection open. Without an explicit close API, callers cannot observe that failure or retry the close after releasing the remaining resources.

**Solution**

The new `Connection.close()` method:

- Calls `sqlite3_close()` synchronously on the connection queue.
- Sets the internal handle to `nil` only when SQLite reports `SQLITE_OK`.
- Throws the SQLite error when close fails, preserving the live handle so callers can continue using the connection or retry closing it later.
- Is idempotent when called after a successful close.
- Is also used from `deinit`, preserving the previous automatic cleanup behavior.

**Tests**

Added regression coverage for the new behavior:

- Verifies that `close()` succeeds and is idempotent.
- Verifies that `close()` throws `SQLITE_BUSY` while a prepared statement is still active.
- Verifies that the connection remains usable after a failed close attempt.
- Verifies that `close()` succeeds after the active statement is released.

### DIFF
--- a/Sources/SQLite/Core/Connection.swift
+++ b/Sources/SQLite/Core/Connection.swift
@@ -161,7 +161,7 @@ public final class Connection {
     }
 
     deinit {
-        sqlite3_close(handle)
+        try? close()
     }
 
     // MARK: -
@@ -184,6 +184,26 @@ public final class Connection {
     /// database via this connection.
     public var totalChanges: Int {
         Int(sqlite3_total_changes(handle))
+    }
+
+    /// Closes the database connection immediately.
+    ///
+    /// Callers should release or exhaust prepared statements, BLOB handles, and backup operations before closing. If
+    /// SQLite reports that the connection is still busy, this method throws and keeps the connection handle open.
+    public func close() throws {
+        try sync {
+            guard let handle = _handle else {
+                return
+            }
+
+            let resultCode = sqlite3_close(handle)
+            if resultCode == SQLITE_OK {
+                _handle = nil
+                return
+            }
+
+            try check(resultCode)
+        }
     }
 
     /// Whether or not the database will return extended error codes when errors are handled.
--- a/Documentation/Index.md
+++ b/Documentation/Index.md
@@ -16,6 +16,7 @@
       - [In-Memory Databases](#in-memory-databases)
       - [URI parameters](#uri-parameters)
       - [Thread-Safety](#thread-safety)
+      - [Closing a Connection](#closing-a-connection)
   - [Building Type-Safe SQL](#building-type-safe-sql)
     - [Expressions](#expressions)
     - [Compound Expressions](#compound-expressions)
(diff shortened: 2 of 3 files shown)


jberkel:
Thanks for the PR! Can you document this change in Documentation/Index.md and rebase the PR against the latest master so that the tests get a chance to run?

RuiNelson:
> Thanks for the PR! Can you document this change in Documentation/Index.md and rebase the PR against the latest master so that the tests get a chance to run?

Done.

I tried to maintain the level of verbosity and tone of the rest of the document.

Sources/SQLite/Core/Connection.swift
```diff
@@ -186,6 +186,26 @@ public final class Connection {
...
+    /// SQLite reports that the connection is still busy, this method throws and keeps the connection handle open.
+    public func close() throws {
+        try sync {
+            guard let handle = _handle else {
+                return
+            }
+
+            let resultCode = sqlite3_close(handle)
+            if resultCode == SQLITE_OK {
+                _handle = nil
```
jberkel, line 201: This will put the `Connection` object in a state where most subsequent operations will fail while force unwrapping the handle (line 88).

The intention of having the `sqlit3_close` happen in the `deinit` was probably to avoid this: either the object is constructed, and usable, or it's no longer in scope (and closed). Being able to close this explicitly puts the connection in a gray zone.

Perhaps the code unwrapping the handle could make this explicit:

```swift
public var handle: OpaquePointer {
    assert(_handle != nil, "connection closed")
    return _handle!
}
```

The code will still fail, but the error message will be clearer.

jberkel:
> > Thanks for the PR! Can you document this change in Documentation/Index.md and rebase the PR against the latest master so that the tests get a chance to run?
> 
> Done.
> 
> I tried to maintain the level of verbosity and tone of the rest of the document.

Sounds like something an AI would say when prompted to write documentation :)

RuiNelson:
> > > Thanks for the PR! Can you document this change in Documentation/Index.md and rebase the PR against the latest master so that the tests get a chance to run?
> > 
> > 
> > Done.
> > I tried to maintain the level of verbosity and tone of the rest of the document.
> 
> Sounds like something an AI would say when prompted to write documentation :)

Actually, it was what I told my AI agent to translate my Portuguese draft to English :o)

Or maybe I'm a machine inside an human-like body that was programmed to think it is human. We'll never know!

jberkel:
> > > > Thanks for the PR! Can you document this change in Documentation/Index.md and rebase the PR against the latest master so that the tests get a chance to run?
> > > 
> > > 
> > > Done.
> > > I tried to maintain the level of verbosity and tone of the rest of the document.
> > 
> > 
> > Sounds like something an AI would say when prompted to write documentation :)
> 
> Actually, it was what I told my AI agent to translate my Portuguese draft to English :o)
> 
> Or maybe I'm a machine inside an human-like body that was programmed to think it is human. We'll never know!

Haha, ok! What do you think about the other suggestions?  A more explicit failure when the connection is closed?

RuiNelson:
I was thinking about throwing an error like, but I need to check the codebase first...

```swift
func getHandle() throws(ConnectionAlreadyClosed) {
 guard let handle else {
  throw ConnectionAlreadyClosed()
 }
 
 return handle
}


// and then, on each method:
let handle = try getHandle()
```

jberkel:
> I was thinking about throwing an error like, but I need to check the codebase first...
> 
> ```swift
> func getHandle() throws(ConnectionAlreadyClosed) {
>  guard let handle else {
>   throw ConnectionAlreadyClosed()
>  }
>  
>  return handle
> }
> 
> 
> // and then, on each method:
> let handle = try getHandle()
> ```

this would be an API change. I'm going to add a precondition in a separate branch.

